PR #1, opened 2018-10-04
Basic waves crypto functions implemented

alexeykiselev:


pkg/crypto/crypto_test.go
```diff
@@ -0,0 +1,358 @@
...
+		expected, err := hex.DecodeString(h)
+		if assert.NoError(t, err) {
+			assert.Equal(t, expected, actual[:])
+		}
+	}
+}
+
+func TestKeccak1(t *testing.T) {
+	const (
+		dataString = "0100000000000000000000000000000000000000000000000000000000000000"
```
montekki: We should move all tests and benchmarks to slices of testcases https://blog.golang.org/subtests

pkg/crypto/crypto.go
```diff
@@ -0,0 +1,176 @@
...
+	h.Write(edPublicKey[:])
+	h.Write(data)
+	h.Sum(hramDigest[:0])
+	var hramDigestReduced [32]byte
+	edwards25519.ScReduce(&hramDigestReduced, &hramDigest)
+
+	var s [32]byte
+	edwards25519.ScMulAdd(&s, &hramDigestReduced, curvePrivateKey, &messageDigestReduced)
+
+	signature = new([64]byte)
```
montekki: Not sure if we need to return a pointer here

pkg/crypto/crypto.go
```diff
@@ -0,0 +1,176 @@
...
+	edPubKeyPoint.ToBytes(edPubKey)
+	signBit := edPubKey[31] & 0x80
+	s := sign(&sk, edPubKey, data)
+	s[63] &= 0x7f
+	s[63] |= signBit
+	copy(sig[:], s[:SignatureSize])
+	return
+}
+
+func sign(curvePrivateKey, edPublicKey *[DigestSize]byte, data []byte) (signature *[SignatureSize]byte) {
```
montekki: Probably named return values are less readable in a large function

pkg/crypto/crypto_test.go
```diff
@@ -0,0 +1,358 @@
+package crypto
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"github.com/mr-tron/base58/base58"
+	"github.com/stretchr/testify/assert"
```
montekki: Let's use the standart test tooling for now to make our code and tests look the same way

PR #4, opened 2018-10-10
Transactions

alexeykiselev:


pkg/proto/types.go
```diff
@@ -0,0 +1,132 @@
...
+	if a.Present {
+		return a.ID.MarshalJSON()
+	}
+	return []byte("null"), nil
+}
+
+// UnmarshalJSON reads Asset from a JSON string Value
+func (a *Asset) UnmarshalJSON(value []byte) error {
+	s := strings.ToUpper(string(value))
+	if s == "NULL" || s == WavesAssetName || s == "" {
```
montekki: You can refactor it with a switch\case https://play.golang.org/p/Y1X8l7cFQfU

alexeykiselev: Cool!

PR #12, opened 2018-10-17
Client

Frozen:


pkg/client/client.go
```diff
@@ -0,0 +1,128 @@
...
+
+type HttpClient interface {
+}
+
+// Creates new client instance
+// If no options provided will use default
+func NewClient(options ...Options) *Client {
+
+	if len(options) > 1 {
+		panic("too many options provided. Expects no or just one item")
```
montekki: please, don't use panics especially in places like a client library

```
panics are always fatal to your program. In panicing you never assume that your caller can solve the problem. Hence panic is only used in exceptional circumstances, ones where it is not possible for your code, or anyone integrating your code to continue.
```
[Dave Chaney](https://dave.cheney.net/2012/01/18/why-go-gets-exceptions-right)

pkg/client/errors.go
```diff
@@ -0,0 +1,25 @@
+package client
```
montekki: Why do we need custom errors here

pkg/client/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+)
+
+// Must create new client or panic
+func mustClient(options ...Options) *Client {
+	client, err := NewClient(options...)
```
montekki: no, this is wrong, should not panic, however the proper error should be returned and in that case the test should fail gracefully

pkg/client/addresses_test.go
```diff
@@ -0,0 +1,167 @@
+package client
+
+import (
+	"context"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/wavesplatform/gowaves/pkg/proto"
+	"os"
+
+	//"os"
```
montekki: This needs to be removed

pkg/client/blocks.go
```diff
@@ -0,0 +1,144 @@
...
+	Generator        string       `json:"generator"`
+	Signature        string       `json:"signature"`
+	Blocksize        uint64       `json:"blocksize"`
+	TransactionCount uint64       `json:"transactionCount"`
+	Fee              uint64       `json:"fee"`
+	//transactions Transaction
+	Height uint64 `json:"height"`
+}
+
+//func (a *Blocks) Signature(ctx context.Context, signature string) (*Block, *Response, error) {
```
montekki: also this needs to be removed

pkg/client/blocks_test.go
```diff
@@ -0,0 +1,109 @@
...
+		client.Blocks.HeadersSeq(context.Background(), 375500, 375500)
+
+	if err != nil {
+		t.Fatalf("expected nil, found %+v", err)
+	}
+	assert.NotNil(t, resp)
+	assert.Equal(t, 1, len(body))
+}
+
+// 2TXfMcQNHJVmkbNoznZrFRLaQHiBayFV9mzxt4VJkyXmxe9aGNn5A2unzUX4M2tqiHEfaWdfCBBo8zJQQpFrCKUY
```
montekki: And this

PR #26, opened 2018-10-23
Feat/montekki/implement protocol connection

montekki:
this is currently a WIP

### DIFF
--- a/pkg/p2p/transport.go
+++ b/pkg/p2p/transport.go
@@ -1,0 +1,21 @@
+package p2p
+
+import (
+	"context"
+	"net"
+	"time"
+)
+
+// DefaultTransport implements the default dialing strategy
+var DefaultTransport = Transport{
+	DialContext: (&net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+		DualStack: true,
+	}).DialContext,
+}
+
+// Transport allows one to reimplement the way we dial a peer
+type Transport struct {
+	DialContext func(ctx context.Context, network, addr string) (net.Conn, error)
+}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -6,7 +6,6 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"net"
 	"time"
@@ -32,14 +31,22 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
 	dialer := net.Dialer{}
 
 	for i := 0; i < 20; i++ {
+		if i > 0 {
+			ticker := time.NewTimer(20 * time.Second)
+
+			select {
+			case <-ticker.C:
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			}
+		}
 		patch = uint32(i)
-		time.Sleep(20 * time.Second)
 		conn, err := dialer.DialContext(ctx, network, addr)
 		if err != nil {
 			continue
 		}
 
-		fmt.Printf("Trying to connect with version %v %v %v", major, minor, patch)
+		zap.S().Infof("Trying to connect with version %v.%v.%v", major, minor, patch)
 		handshake := proto.Handshake{Name: "wavesT",
 			Version:           proto.Version{Major: major, Minor: minor, Patch: patch},
 			NodeName:          "gowaves",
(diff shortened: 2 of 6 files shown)


pkg/server/server.go
```diff
@@ -20,35 +25,64 @@ type Server struct {
...
-		zap.S().Error("failed to send handshake: ", err)
-		return
-	}
-	_, err = handshake.ReadFrom(conn)
-	if err != nil {
-		zap.S().Error("failed to read handshake: ", err)
-		return
+func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	var major, minor, patch uint32
+	minor = 5
```
kazhuravlev: this is a constant. don`t you think so?

pkg/server/server.go
```diff
@@ -20,35 +25,64 @@ type Server struct {
...
+
+	for i := 0; i < 20; i++ {
+		patch = uint32(i)
+		time.Sleep(20 * time.Second)
+		conn, err := dialer.DialContext(ctx, network, addr)
+		if err != nil {
+			continue
+		}
+
+		fmt.Printf("Trying to connect with version %v %v %v", major, minor, patch)
```
kazhuravlev: will be better if you using a logger instead fmt

pkg/server/server.go
```diff
@@ -20,35 +25,64 @@ type Server struct {
...
-		return
+func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	var major, minor, patch uint32
+	minor = 5
+
+	dialer := net.Dialer{}
+
+	for i := 0; i < 20; i++ {
+		patch = uint32(i)
+		time.Sleep(20 * time.Second)
```
kazhuravlev:
```suggestion
		ticker := time.NewTimer(20*time.Second)
		
		select {
		case <- ticker.C:
		case <-ctx.Done():
		    return nil, ctx.Err()
		}
```

pkg/server/server.go
```diff
@@ -20,35 +25,64 @@ type Server struct {
...
 	}
+	return nil, errors.New("TODO")
+}
 
-	var b []byte
-	b, e := json.Marshal(handshake)
-	if e != nil {
+func handleClient(peer string) {
+	customTransport := p2p.Transport{DialContext: dialContext}
+	conn, _ := p2p.NewConn(
```
kazhuravlev: handle `err`

pkg/server/server.go
```diff
@@ -20,35 +25,64 @@ type Server struct {
...
+		p2p.WithVersion(proto.Version{Major: 0, Minor: 5, Patch: 14}),
```
kazhuravlev: repeated constant

pkg/server/server.go
```diff
@@ -130,13 +164,9 @@ func (m *Server) Run() {
 }
 
 func (m *Server) RunClients() {
```
kazhuravlev: Must receive `context.Context`

pkg/server/server.go
```diff
@@ -130,13 +164,9 @@ func (m *Server) Run() {
...
+
 	for _, peer := range m.BootPeerAddrs {
-		conn, err := net.Dial("tcp", peer)
-		if err != nil {
-			zap.S().Error("failed to connect to peer: ", err)
-			continue
-		}
 
-		go handleClient(conn)
+		go handleClient(peer)
```
kazhuravlev: must receive `context.Context`

pkg/p2p/transport.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	"net"
+	"time"
+)
+
+// DefaultTransport implements the default dialing strategy
+var DefaultTransport = Transport{
+	DialContext: (&net.Dialer{
+		Timeout:   30 * time.Second,
```
kazhuravlev: hardcode

pkg/p2p/transport.go
```diff
@@ -0,0 +1,21 @@
+package p2p
+
+import (
+	"context"
+	"net"
+	"time"
+)
+
+// DefaultTransport implements the default dialing strategy
+var DefaultTransport = Transport{
```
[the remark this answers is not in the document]
montekki: yes but it doesn't live in it's own ```transport``` package, but in the ```p2p```

pkg/server/server.go
montekki: no, the idea behind this is following: there is a default ```DialContext``` implementation in the transport that just wraps the ```net.Dialer.DialContext``` and tries to connect to a node.
Here we redefine the ```DefaultTransport.DialContext``` behaviour to try to re-adjust the connection number of the node. So, for instance, if the version of the node we are trying to connect to is ```0.5.5``` we begin with trying to connect with version ```0.5.10```, the connection gets terminated, we try ```0.5.9``` and so on. Most likely we will have to connect to a large variety of combinations of Major.Minor.Patch versions. Up until this day only the Major version is constant and for every existing node in the world is ```0```.

PR #27, opened 2018-10-24
GO-3: DataV1 transaction implemented

alexeykiselev:


pkg/proto/common.go
```diff
@@ -25,6 +25,28 @@ func StringWithUInt16Len(buf []byte) (string, error) {
...
+//PutBytesWithUInt16Len prepends given buf with 2 bytes of it's length.
+func PutBytesWithUInt16Len(buf []byte, data []byte) {
+	sl := uint16(len(data))
+	binary.BigEndian.PutUint16(buf, sl)
+	copy(buf[2:], data)
+}
+
+// BytesWithUInt16Len reads from buf an array of bytes of length encoded in first 2 bytes.
+func BytesWithUInt16Len(buf []byte) ([]byte, error) {
+	if l := len(buf); l < 2 {
```
montekki: Is there a reason not simplify this and the following condition?

PR #30, opened 2018-10-25
client peers implementation

Frozen:


pkg/client/client.go
```diff
@@ -70,6 +71,7 @@ func NewClient(options ...Options) (*Client, error) {
 		Blocks:    NewBlocks(opts),
 		Wallet:    NewWallet(opts),
 		Alias:     NewAlias(opts),
+		Peers:     NewPeers(opts),
```
montekki: Why the generic config options pattern is not implemented here?

Frozen: Such pattern adds and removes functionality depends on passed params, in my case i just initialize public property. I think my version is much clean and obvious.

PR #59, opened 2018-11-29
command line interface wallet and sign

Frozen:


cmd/sign/main.go
```diff
@@ -0,0 +1,179 @@
...
+		Recipient    string
+		Amount       uint64
+		Fee          uint64
+		CustomSecret string
+	}
+
+	opts := Opts{}
+
+	f := flag.NewFlagSet("transfer", flag.ExitOnError)
+	f.StringVarP(&opts.PathToWallet, "path", "p", "", "Path to wallet")
```
alexeykiselev: Isn't better "wallet" and "w"?

pkg/wallet/crypt.go
```diff
@@ -0,0 +1,68 @@
...
+	"crypto/sha256"
+	"fmt"
+	"io"
+)
+
+type crypt struct {
+	key []byte
+}
+
+func NewCrypt(key []byte) *crypt {
```
alexeykiselev: I suggest to use https://godoc.org/golang.org/x/crypto/argon2 for password hashing instead of hash5000

PR #73, opened 2019-02-11
Synchronizer's binary search of last common block fixed and simplified

alexeykiselev:


cmd/wmd/internal/synchronizer.go
```diff
@@ -74,6 +74,9 @@ func (s *Synchronizer) run() {
 			s.ticker.Stop()
 			return
 		case <-s.ticker.C:
+			if s.interrupted() {
```
Frozen: you previously check "case <-s.interrupt:" and call s.ticker.Stop(), then check s.interrupted again
is it correct?

alexeykiselev: Yes, looks like overkill, but it allows to stop sync after tick event happened first.

PR #78, opened 2019-02-14
State fixes and improvements

zer0main:


pkg/storage/accountsstorage_test.go
```diff
@@ -15,37 +15,21 @@ const (
 	TOTAL_BLOCKS_NUMBER = 200
```
alexeykiselev: Please use **camelCase** or **UpperCamelCase** for constants.

pkg/importer/importer.go
```diff
@@ -15,19 +15,20 @@ import (
...
 }
 
-func ApplyFromFile(st State, blockchainPath string, nBlocks int, checkBlocks bool) error {
+func ApplyFromFile(st State, blockchainPath string, nBlocks, startHeight uint64, checkBlocks bool) error {
 	blockchain, err := os.Open(blockchainPath)
 	if err != nil {
 		return errors.Errorf("failed to open blockchain file: %v\n", err)
 	}
 	sb := make([]byte, 4)
 	buf := make([]byte, 2*1024*1024)
```
alexeykiselev: Better create not slice of bytes, but fixed byte array of the maximum block size. And later create a slice of required length from it.

zer0main:
> I think that `keys.go` should be not in `proto` package but in `state` package.

The problem is that `proto` package uses keys too, and `state` already imports `proto`.

pkg/importer/importer.go
zer0main: Done.

PR #84, opened 2019-03-04
Clarify internal (state) API, add custom error types

zer0main:


zer0main:
Check out `state/api.go` for current state's API overview.

pkg/state/api.go
```diff
@@ -0,0 +1,93 @@
...
+		return 0
+	}
+}
+
+// State represents overall Node's state.
+// Data retrievals (e.g. account balances), as well as modifiers (like adding or rolling back blocks)
+// should all be made using this interface.
+type State interface {
+	// Block getters.
+	GetBlock(blockID crypto.Signature) (*proto.Block, error)
```
Frozen: remove "Get" prefix from methods
https://golang.org/doc/effective_go.html#Getters

pkg/state/accountsstorage.go
```diff
@@ -128,14 +128,14 @@ func (s *localStor) reset() {
 	s.assets = make(map[assetBalanceKey][]byte)
 }
 
-type ID2Height interface {
-	HeightByBlockID(blockID crypto.Signature) (uint64, error)
+type iD2Height interface {
```
alexeykiselev: May be `idToHeight` better?

pkg/state/api.go
zer0main: Done.

pkg/state/accountsstorage.go
zer0main: Done.

PR #85, opened 2019-03-05
State: support blocks score

zer0main:


pkg/state/keys.go
```diff
@@ -75,3 +80,13 @@ func (k *txOffsetKey) bytes() []byte {
...
 }
+
+type scoreKey struct {
+	height uint64
+}
+
+func (k *scoreKey) bytes() []byte {
+	heightBuf := make([]byte, 8)
+	binary.LittleEndian.PutUint64(heightBuf, k.height)
+	return append([]byte{scoreKeyPrefix}, heightBuf...)
```
alexeykiselev: I think, copy is faster, better not to use append

pkg/state/score.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/pkg/errors"
+	"github.com/wavesplatform/gowaves/pkg/keyvalue"
+)
+
+func calculateScore(baseTarget uint64) (*big.Int, error) {
+	if baseTarget == 0 {
+		return nil, errors.New("zero base target")
+	}
+	var max big.Int
+	max.SetString("18446744073709551616", 10)
```
alexeykiselev: Maybe extract `max` to global var? Is it possible initialise it not from string?

zer0main: It's only possible to initialize it from string or from bytes slice, since it's bigger than `MaxUint64`.

pkg/state/keys.go
zer0main: I can actually `PutUint64` directly into buf sub-slice, fixed.

zer0main:
Ready to merge.

PR #86, opened 2019-03-19
Complete block headers validation

zer0main:


pkg/consensus/blocks_validation.go
```diff
@@ -0,0 +1,250 @@
...
+	if header.Version == proto.GenesisBlockVersion || header.Version == proto.PlainBlockVersion {
+		return nil
+	}
+	if height < cv.settings.BlockVersion3AfterHeight {
+		return errors.Errorf("block version 3 can only appear after %d height", cv.settings.BlockVersion3AfterHeight)
+	}
+	return nil
+}
+
+func (cv *ConsensusValidator) checkTargetLimit(height, target uint64) error {
```
alexeykiselev: I think this feature will be activated by voting. It's very unstable check, could play bad on old blocks. Is it unlimited in config?

pkg/settings/blockchain_settings.go
```diff
@@ -0,0 +1,78 @@
...
+		TxFromFutureCheckAfterTime:             1479168000000,
+		MinimalGeneratingBalanceCheckAfterTime: 1479168000000,
+
+		MaxTxTimeBackOffset:    120 * 60000,
+		MaxTxTimeForwardOffset: 90 * 60000,
+
+		AddressSchemeCharacter: proto.MainNetScheme,
+
+		AverageBlockDelaySeconds: 60,
+		MaxBaseTarget:            200,
```
alexeykiselev: For example, BT at height 200 is 641551570. 
Or this setting works only for FairPOS?

zer0main: Only for Fair PoS.

PR #89, opened 2019-03-29
Split transactions by versions in proto, separate common parts and make them accessible

zer0main:


pkg/proto/types.go
```diff
@@ -295,7 +295,26 @@ type Order interface {
...
+		if !ok {
+			return OrderData{}, errors.New("failed to cast an order version 2 to OrderV2")
+		}
+		return o.OrderData, nil
+	default:
+		return OrderData{}, errors.New("invalid order version")
+	}
+}
+
+type OrderData struct {
```
alexeykiselev: I suggest to name this structure 'OrderBody', or may be we should rename Order interface...

zer0main: Done.

PR #90, opened 2019-04-05
Implement validation against state for: Issue, Reissue, Burn, Exchange txs

zer0main:


pkg/proto/transactions.go
```diff
@@ -455,7 +455,7 @@ func (tx *Transfer) unmarshalBinary(data []byte) error {
 }
 
 type Issue interface {
-	GetID() *crypto.Digest
+	GetID() []byte
```
alexeykiselev: Is it necessary? Now you have to check not only emptiness, but the size too...

pkg/proto/transactions.go
```diff
@@ -526,10 +526,10 @@ func (tx *Reissue) unmarshalBinary(data []byte) error {
 }
 
 type Exchange interface {
-	GetID() *crypto.Digest
+	GetID() []byte
```
alexeykiselev: The same question as above.

pkg/proto/transactions.go
```diff
@@ -526,10 +526,10 @@ func (tx *Reissue) unmarshalBinary(data []byte) error {
...
 	GetSenderPK() crypto.PublicKey
-	GetBuyOder() OrderBody
-	GetSellOrder() OrderBody
+	GetBuyOrder() (OrderBody, error)
```
alexeykiselev: Why error here? What could go wrong?

PR #102, opened 2019-06-07
Go 126 apply and rollback

Frozen:


pkg/mainer/keypair.go
```diff
@@ -0,0 +1,25 @@
+package mainer
```
alexeykiselev: Typo? Shouldn't be 'miner'?

pkg/mainer/mainer.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/wavesplatform/gowaves/pkg/proto"
+	"github.com/wavesplatform/gowaves/pkg/state"
+	"github.com/wavesplatform/gowaves/pkg/types"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+
+	"time"
+)
+
+type Mainer struct {
```
alexeykiselev: Miner?

pkg/mainer/keypair.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/wavesplatform/gowaves/pkg/crypto"
+	"github.com/wavesplatform/gowaves/pkg/proto"
+)
+
+type KeyPair struct {
```
alexeykiselev: Not implemented? Looks like a duplication of new proto.KeyPair...

PR #145, opened 2019-08-30
verify function

Frozen:


pkg/types/types.go
```diff
@@ -40,3 +40,18 @@ type TransactionWithBytes struct {
 	T proto.Transaction
 	B []byte
 }
+
+// state for smart contracts
+type SmartState interface {
+	NewLastHeight() (uint64, error)
```
zer0main: `NewestHeight()`

pkg/types/types.go
```diff
@@ -40,3 +40,18 @@ type TransactionWithBytes struct {
...
+	TransactionByID([]byte) (proto.Transaction, error)
+	TransactionHeightByID([]byte) (uint64, error)
+	Account(proto.Recipient) Account
```
zer0main: `AccountBalance(addr proto.Address, asset []byte) (uint64, error)`

pkg/types/types.go
```diff
@@ -40,3 +40,18 @@ type TransactionWithBytes struct {
...
+}
+
+// Some abstract way to get account related info
+type Account interface {
+	Data() []proto.DataEntry
```
zer0main: It isn't needed anywhere to retrieve all data entries at once. It can be heavy data.
Instead, we already have the following methods (`state/api.go`):
 ```
        // Accounts data storage.
        RetrieveNewestEntry(addr proto.Address, key string) (proto.DataEntry, error)
        RetrieveNewestIntegerEntry(addr proto.Address, key string) (*proto.IntegerDataEntry, error)
        RetrieveNewestBooleanEntry(addr proto.Address, key string) (*proto.BooleanDataEntry, error)
        RetrieveNewestStringEntry(addr proto.Address, key string) (*proto.StringDataEntry, error)
        RetrieveNewestBinaryEntry(addr proto.Address, key string) (*proto.BinaryDataEntry, error)
```

File: pkg/ride/evaluator/evaluate/evaluate.go [hunk 4]
```diff
@@ -22,3 +24,22 @@ func Eval(e ast.Expr, s ast.Scope) (bool, error) {
 
 	return b.Value, nil
 }
+
+func Verify(scheme byte, state types.SmartState, script *ast.Script, transaction proto.Transaction) (bool, error) {
```
zer0main: `scheme` can be retrieved using `state.BlockchainSettings().AddressSchemeCharacter`, maybe we shouldn't add extra argument here.

zer0main: Still applies.

zer0main: Still applies.

PR #190, opened 2019-10-28
Suspend nodes and pass peers as reference

Frozen:


cmd/node/node.go
```diff
@@ -70,7 +71,12 @@ func main() {
 		return
 	}
 
-	state, err := state.NewState("./", state.DefaultStateParams(), settings.MainNetSettings)
+	cfg := settings.MainNetSettings
+	if cli.Run.GenesisPath != "" {
+		cfg.GenesisGetter = settings.FromPath(cli.Run.GenesisPath)
```
zer0main: If there is genesis block different from MainNet, then the rest of the settings (`cfg`) should be either `Testnet` or `Custom`. 
MainNet settings won't work for Testnet blockchain, and likely won't suit Custom blockchain.
We should either disable ability to switch genesis block path, or enable ability to parse custom settings from config.

PR #263, opened 2020-03-13
Transactions merkle root hash

### LINKED ISSUE #228: [V1.2] Add MerkleRootHash of transactions
**Аbstract**
Add MerkleRootHash of block's transactions. 

**Motivation and Purposes**
Block generator signs only block header which in turn contains a root hash of Merkle tree of all block's transactions. This change allows to verify block without downloading all block's transactions and will be used to develop sidechains.

**Specification**

The Merkle tree should be build as follows:

1. Calculate the digests of all block's transactions: hash(L1), hash(L2), hash(L3) and so on.
2. Calculate the digest of pairs of transaction's digests. For example, hash(hash(L1) + hash(L2)). If block contains odd number of transactions, the last transaction digest is used twice to calculate the pair's digest: hash (hash(L3) + hash(L3)).
3. Repeat the step 2 until only one digest left, this is the root hash.
4. Save the root hash in block's header.

```
hash (hash (hash(L1) + hash(L2)) + hash (hash(L3) + hash(L3))) = Root Hash
```

Notes:

1. Transaction's digest is calculated from whole transaction including signature.
2. Blake2b256 hash function should be used to calculate digests in Merkle tree.

alexeykiselev:
Closes #228


pkg/miner/miner.go
```diff
@@ -238,9 +238,15 @@ func (a *MicroblockMiner) mineMicro(ctx context.Context, rest restLimits, blockA
 	}
 
 	priv := keyPair.Secret
+	if newBlock.Version >= proto.ProtoBlockVersion {
+		err = newBlock.SetTransactionsRoot(a.scheme)
```
zer0main: Can we move `SetTransactionRoot()` call to `CreateBlock()` function?
It is used in multiple places, so might be better this way.

pkg/miner/internal.go
```diff
@@ -10,6 +10,12 @@ func MineBlock(version proto.BlockVersion, nxt proto.NxtConsensus, pair proto.Ke
 	if err != nil {
 		return nil, err
 	}
+	if b.Version >= proto.ProtoBlockVersion {
+		err := b.SetTransactionsRoot(scheme)
```
zer0main: Can we move `SetTransactionRoot()` call to `CreateBlock()` function?
It is used in multiple places, so might be better this way.

pkg/state/block_differ_test.go
```diff
@@ -38,7 +38,7 @@ func genBlocks(t *testing.T, to *blockDifferTestObjects) (*proto.Block, *proto.B
 	txs := proto.Transactions{createTransferWithSig(t)}
 	randSig := genRandBlockIds(t, 1)[0]
 	genSig := crypto.MustBytesFromBase58(defaultGenSig)
-	parent, err := proto.CreateBlock(txs, 1565694219644, randSig, testGlobal.matcherInfo.pk, proto.NxtConsensus{BaseTarget: 65, GenSignature: genSig}, proto.NgBlockVersion, nil, -1)
```
zer0main: `TestNetScheme` -> `MainNetScheme`. This block is not Protobuf block, so scheme does not really affect anything, but it is then Signed() using `MainNetScheme`, so it should be the same for clarity.

pkg/state/block_differ_test.go
```diff
@@ -47,7 +47,7 @@ func genBlocks(t *testing.T, to *blockDifferTestObjects) (*proto.Block, *proto.B
 	txs = []proto.Transaction{createIssueWithSig(t, 1000)}
 	genSig, err = to.gsp.GenerationSignature(testGlobal.minerInfo.pk, parent.GenSignature[:])
 	require.NoError(t, err, "GeneratorSignature() failed")
-	child, err := proto.CreateBlock(txs, 1565694219944, parent.BlockSignature, testGlobal.minerInfo.pk, proto.NxtConsensus{BaseTarget: 66, GenSignature: genSig}, proto.NgBlockVersion, nil, -1)
+	child, err := proto.CreateBlock(txs, 1565694219944, parent.BlockSignature, testGlobal.minerInfo.pk, proto.NxtConsensus{BaseTarget: 66, GenSignature: genSig}, proto.NgBlockVersion, nil, -1, proto.TestNetScheme)
```
zer0main: `TestNetScheme` -> `MainNetScheme`.

pkg/state/block_differ_test.go
```diff
@@ -173,7 +173,7 @@ func genBlockWithSingleTransaction(t *testing.T, prevID crypto.Signature, prevGe
 	txs := proto.Transactions{genTransferWithWavesFee(t)}
 	genSig, err := to.gsp.GenerationSignature(testGlobal.minerInfo.pk, prevGenSig)
 	require.NoError(t, err)
-	block, err := proto.CreateBlock(txs, 1565694219944, prevID, testGlobal.minerInfo.pk, proto.NxtConsensus{BaseTarget: 66, GenSignature: genSig}, proto.RewardBlockVersion, nil, -1)
+	block, err := proto.CreateBlock(txs, 1565694219944, prevID, testGlobal.minerInfo.pk, proto.NxtConsensus{BaseTarget: 66, GenSignature: genSig}, proto.RewardBlockVersion, nil, -1, proto.TestNetScheme)
```
zer0main: `TestNetScheme` -> `MainNetScheme`.

pkg/state/block_differ_test.go
```diff
@@ -38,7 +38,7 @@ func genBlocks(t *testing.T, to *blockDifferTestObjects) (*proto.Block, *proto.B
 	txs := proto.Transactions{createTransferWithSig(t)}
 	randSig := genRandBlockIds(t, 1)[0]
 	genSig := crypto.MustBytesFromBase58(defaultGenSig)
-	parent, err := proto.CreateBlock(txs, 1565694219644, randSig, testGlobal.matcherInfo.pk, proto.NxtConsensus{BaseTarget: 65, GenSignature: genSig}, proto.NgBlockVersion, nil, -1)
+	parent, err := proto.CreateBlock(txs, 1565694219644, randSig, testGlobal.matcherInfo.pk, proto.NxtConsensus{BaseTarget: 65, GenSignature: genSig}, proto.NgBlockVersion, nil, -1, proto.TestNetScheme)
```
zer0main: `TestNetScheme` -> `MainNetScheme`. This block is not Protobuf block, so scheme does not really affect anything, but it is then Signed() using `MainNetScheme`, so it should be the same for clarity.

alexeykiselev: Actually it's very strange why we use MainNet in tests :)

PR #272, opened 2020-03-17
Exchange orders any order

### LINKED ISSUE #266: [V1.2] Support any order of orders in Exchange transaction
**Аbstract**
Now it's validated that the first order in an exchange transaction have to be a BUY order and the second is SELL. Protobuf version of Exchange transactions should support any order of orders in transaction.

alexeykiselev:
Closes #266

### DIFF
--- a/pkg/state/transaction_checker.go
+++ b/pkg/state/transaction_checker.go
@@ -588,37 +588,48 @@ func (tc *transactionChecker) checkExchange(transaction proto.Transaction, info 
 	if err := tc.checkTimestamps(tx.GetTimestamp(), info.currentTimestamp, info.parentTimestamp); err != nil {
 		return nil, errors.Wrap(err, "invalid timestamp")
 	}
-	if err := tc.checkEnoughVolume(tx.GetSellOrderFull(), tx.GetSellMatcherFee(), tx.GetAmount(), info); err != nil {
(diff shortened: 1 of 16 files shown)


pkg/state/transaction_checker.go
```diff
@@ -588,37 +588,53 @@ func (tc *transactionChecker) checkExchange(transaction proto.Transaction, info
...
+	if err != nil {
+		return nil, err
 	}
-	if err := tc.checkEnoughVolume(tx.GetBuyOrderFull(), tx.GetBuyMatcherFee(), tx.GetAmount(), info); err != nil {
-		return nil, errors.Wrap(err, "exchange transaction; buy order")
+	if tx.GetOrder1().GetOrderType() != proto.Buy && tx.GetOrder2().GetOrderType() != proto.Sell {
+		if !proto.IsProtobufTx(transaction) {
+			return nil, errors.New("sell order not allowed on first place in exchange transaction of versions prior 4")
+		}
+		if !blockV5Activated {
```
zer0main: Protobuf transactions are not allowed before blockV5 activation [by appender](https://github.com/wavesplatform/gowaves/blob/07e4ca98dcd36bb748d67fdb51890d8b25a1b0cd/pkg/state/appender.go#L234), so this check is not needed.

PR #275, opened 2020-03-24
Use header hash as block ID for Protobuf blocks

### LINKED ISSUE #265: [V1.2] Replace BlockSignature with hash digest of block header as Block ID
**Аbstract**
Currently the block signature is used as a block ID to reference to. It should be replaced with hash digest of block's header. 

** Specification**

- [x] Replace block signature with hash digest of block's header
- [x] Add new network messages to request and receive blocks IDs

zer0main:
Closes #265.


pkg/proto/block.go
```diff
@@ -34,23 +36,133 @@ func (nc *NxtConsensus) BinarySize() int {
...
+)
+
+type BlockID struct {
+	sig    crypto.Signature
+	dig    crypto.Digest
+	idType BlockIDType
+}
+
+func NewBlockIDFromBase58(s string) (BlockID, error) {
+	dig, err := crypto.NewDigestFromBase58(s)
```
alexeykiselev: Is it correct to try to created a digest, that is shorter, first? It seems like we can build digest from any signature...

alexeykiselev: May be it is better to convert string to bytes and then use `NewBlockIDFromBytes`?

zer0main: It checks that decoded length is exactly 32, so it shouldn't accept signatures.

zer0main: It checks that decoded length is exactly 32, so it shouldn't accept signatures. Internally, it converts string to bytes and checks its length for both `Digest` and `Signature`.

PR #289, opened 2020-04-15
Extended api reimplemented with new runtime logic

Frozen:


pkg/grpc/client_test.go
```diff
@@ -42,6 +42,35 @@ func TestTransactionsAPIClient(t *testing.T) {
 	assert.Equal(t, io.EOF, err)
 }
 
+func TestTransactions_GetTransactions(t *testing.T) {
+	t.SkipNow()
```
zer0main, line 46: Why do we always skip this test? Should we remove it?

pkg/proto/address.go
```diff
@@ -38,11 +38,8 @@ const (
 // Address is the transformed Public Key with additional bytes of the version, a blockchain scheme and a checksum.
 type Address [AddressSize]byte
 
-func (a Address) Body() ([]byte, error) {
-	if len(a) != AddressSize {
-		return nil, errors.New("invalid address length")
-	}
-	return a[headerSize : headerSize+bodySize], nil
+func (a Address) Body() []byte {
+	return a[headerSize : headerSize+bodySize]
```
zer0main, line 42: Are you sure it is safe to do it? This function is called in many other functions, are we sure
that data never comes from user? If there is at least one such place, then this is a vulnerability
which might allow external users to stop the node by `panic`ing here.

PR #292, opened 2020-04-16
State hashes

zer0main:


pkg/util/common/util.go
```diff
@@ -152,6 +153,42 @@ func ParseDuration(str string) (uint64, error) {
 	return total, nil
 }
 
+func ToBase64JSON(b []byte) []byte {
```
alexeykiselev, line 156: Isn't that a default behaviour for byte arrays JSON serialization?

PR #294, opened 2020-04-20
Switch to using submodule for schemas

zer0main:


Makefile
```diff
@@ -154,3 +154,6 @@ mock:
 	mockgen -source pkg/p2p/peer/peer.go -destination pkg/mock/peer.go -package mock Peer
 	mockgen -source pkg/state/api.go -destination pkg/mock/state.go -package mock State
 	mockgen -source pkg/node/state_fsm/default.go -destination pkg/node/state_fsm/default_mock.go -package state_fsm Default
+
+proto:
+	./genpb.sh
```
alexeykiselev, line 159: Let's move commands for generation here and remove `genpb.sh`. It will be the same as for mocks generation. Later we can introduce a `generate` command that do both.

genpb.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
alexeykiselev, line 1: Move those commands to `Makefile`

pkg/grpc/README.md
```diff
@@ -12,20 +12,6 @@ go get -u github.com/golang/protobuf/protoc-gen-go
 
 ## Package structure
 
-* `grpc/proto/` - a copy of proto files from [protobuf-schemas](https://github.com/wavesplatform/protobuf-schemas) project. Files are copied from folders `proto/waves/` and `proto/waves/node/grpc`. And `import` directives updated afterwards to reflect the flat structure.
+* `grpc/protobuf-schemas/` - a submodule of [protobuf-schemas](https://github.com/wavesplatform/protobuf-schemas) project (proto files).
 * `grpc/generated` - code generated from proto files.
 * `grpc/server` - gRPC server implementation (API).
```
alexeykiselev, line 17: It will be useful to put here information how to update submodule and regenerate protobuf code.

PR #295, opened 2020-04-21
Microblock generation and apply with vrf.

Frozen:


pkg/node/messages/broadcase_transaction.go
```diff
@@ -0,0 +1,15 @@
+package messages
+
+import "github.com/wavesplatform/gowaves/pkg/proto"
+
+type BroadcastTransaction struct {
```
alexeykiselev, line 5: Typo in file name `broadcase_transaction.go` -> `broadcast_transactions.go`

Issue #303, opened 2020-05-06
[V1.2][RIDE V4] Function assetBalance should fail on unit assetId


alexeykiselev:
**Specification**

Function `assetBalance(addressOrAlias: Address|Alias, assetId: ByteVector): Int` should fail in case of Unit is passed as `assetId`.

PR #312, opened 2020-05-08
Function to get waves balance details from state implemented

### LINKED ISSUE #302: [V1.2][RIDE V4] Update wavesBalance function
**Specification**

In RIDE v4 function `wavesBalance(Address|Alias): BalanceDetails` should return `BalanceDetails` structure. `BalanceDetails` structure consists of 4 fields:

 * `regular` - amount of Waves
 * `available` - amount of Waves that can be spent (except leased Waves)
 * `effective` - amount of Waves including leased Waves
 * `generating` - minimal amount of `effective` balance during 1000 last blocks

alexeykiselev:
Closes #302, #303

### DIFF
--- a/pkg/mock/state.go
+++ b/pkg/mock/state.go
@@ -250,21 +250,6 @@ func (mr *MockStateInfoMockRecorder) FullWavesBalance(account interface{}) *gomo
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "FullWavesBalance", reflect.TypeOf((*MockStateInfo)(nil).FullWavesBalance), account)
 }
 
-// NewestFullWavesBalance mocks base method
-func (m *MockStateInfo) NewestFullWavesBalance(account proto.Recipient) (*proto.FullWavesBalance, error) {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "NewestFullWavesBalance", account)
-	ret0, _ := ret[0].(*proto.FullWavesBalance)
-	ret1, _ := ret[1].(error)
-	return ret0, ret1
-}
-
-// NewestFullWavesBalance indicates an expected call of NewestFullWavesBalance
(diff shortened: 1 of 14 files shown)


pkg/state/threadsafe_wrapper.go
```diff
@@ -85,6 +85,12 @@ func (a *ThreadSafeReadWrapper) FullWavesBalance(account proto.Recipient) (*prot
 	return a.s.FullWavesBalance(account)
 }
 
+func (a *ThreadSafeReadWrapper) NewestFullWavesBalance(recipient proto.Recipient) (*proto.FullWavesBalance, error) {
```
zer0main, line 88: `Newest` functions should not be added here.

pkg/state/api.go
```diff
@@ -43,6 +43,7 @@ type StateInfo interface {
 	HeightToBlockID(height proto.Height) (proto.BlockID, error)
 	// FullWavesBalance returns complete Waves balance record.
 	FullWavesBalance(account proto.Recipient) (*proto.FullWavesBalance, error)
+	NewestFullWavesBalance(account proto.Recipient) (*proto.FullWavesBalance, error)
```
zer0main, line 46: `Newest` functions should not be added to this interface. They are not thread safe, they are only intended to be used by `ride` and `consensus` packages, not the node or api.

pkg/mock/state.go
```diff
@@ -250,6 +250,21 @@ func (mr *MockStateInfoMockRecorder) FullWavesBalance(account interface{}) *gomo
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "FullWavesBalance", reflect.TypeOf((*MockStateInfo)(nil).FullWavesBalance), account)
 }
 
+// NewestFullWavesBalance mocks base method
+func (m *MockStateInfo) NewestFullWavesBalance(account proto.Recipient) (*proto.FullWavesBalance, error) {
```
zer0main, line 254: Should be removed from mock also (see below).

PR #322, opened 2020-05-29
Fix stagenet v1.2.5

alexeykiselev:


pkg/node/actions_by_type.go
```diff
@@ -177,7 +177,7 @@ func PBBlockAction(_ services.Services, mess peer.ProtoMessage, fsm state_fsm.FS
 		zap.S().Debug(err)
 		return fsm, nil, err
 	}
-	zap.S().Info("PBBlockAction ", b.BlockSignature)
+	zap.S().Debugf("Protobuf block received '%s'", b.ID)
```
zer0main, line 180: Should call `b.ID.String()` here to get string.

pkg/proto/protobuf_converters.go
```diff
@@ -351,7 +351,6 @@ func (c *ProtobufConverter) attachment(att *g.Attachment, untyped bool) Attachme
 		return nil
 	}
 	if att == nil {
-		c.err = errors.New("empty attachment")
```
zer0main, line 354: In `pkg/ride/evaluator/ast/transactions.go` we use tx.Attachment. We should modify code to check Attachments for nil there.

PR #327, opened 2020-06-22
Simplify blocks addition and rollback; extend UTX validation

zer0main:


pkg/state/batched_storage.go
```diff
@@ -463,7 +463,9 @@ func (s *batchedStorage) newBackwardRecordIterator(key []byte) (*recordIterator,
 	return newRecordIterator(batchIter, s.params.recordSize), nil
 }
 
-func (s *batchedStorage) normalize(batch []byte) ([]byte, error) {
+type blockValidationFunc func(blockNum uint32) (bool, error)
```
Frozen, line 466: maybe better 
`type blockValidationFunc func(blockNum uint32) error`

PR #329, opened 2020-07-07
Untyped attachments

Frozen:


cmd/node/node.go
```diff
@@ -319,6 +319,8 @@ func main() {
 	n := node.NewNode(services, declAddr, bindAddr, proto.NewTimestampFromUSeconds(outdatePeriod))
 	go n.Run(ctx, parent, services.InternalChannel)
 
+	go sched.Reschedule()
```
alexeykiselev, line 322: What is the point of doing reschedule here? There is no connection and no consensus at this moment. I think this is pointless.

pkg/ride/evaluator/reader/reader.go
```diff
@@ -148,6 +148,10 @@ func (a *BytesReader) ReadBytes() []byte {
 	return out
 }
 
+func (a *BytesReader) Content() []byte {
```
alexeykiselev, line 151: No usages

cmd/node/node.go
Frozen: Node can mining without connections. There should be initial schedule request.

pkg/ride/evaluator/reader/reader.go
Frozen: Its used by foreign project.

PR #334, opened 2020-07-22
Leasing tests.

Frozen:


pkg/grpc/server/accounts_api.go
```diff
@@ -82,13 +83,25 @@ func (s *Server) GetActiveLeases(req *g.AccountRequest, srv g.AccountsApi_GetAct
...
-	reqTr := &g.TransactionsRequest{Sender: req.Address}
-	ftr, err := newTxFilter(s.scheme, reqTr)
+	var c proto.ProtobufConverter
+	addr, err := c.Address(s.scheme, req.Address)
 	if err != nil {
-		return status.Errorf(codes.FailedPrecondition, err.Error())
+		return status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	filterFn := func(tx proto.Transaction) bool {
```
zer0main, line 92: Can we implement another `filter` struct which would filter by `sender` OR `recipient` instead of `sender` AND `recipient`?
Code would look better if we use the same mechanism everywhere.
Otherwise we need to remove `txFilerLeases`, it is not used anymore.

PR #339, opened 2020-07-30
Invoke script transaction tests.

Frozen:


pkg/state/invoke_applier.go
```diff
@@ -252,6 +265,10 @@ func (ia *invokeApplier) fallibleValidation(tx *proto.InvokeScriptWithProofs, in
 	if err := ia.saveIntermediateDiff(totalChanges.diff); err != nil {
 		return proto.DAppError, info.failedChanges, err
 	}
+	// Empty keys rejected since protobuf version.
+	if ia.countEmptyDataEntryKeys(info.actions) > 0 {
```
zer0main, line 269: `if proto.IsProtobufTx(tx) && ...`
Specification says this check is only relevant for Protobuf transactions (Invoke version >= 2).

PR #348, opened 2020-08-20
Api update.

Frozen:


pkg/client/blocks.go
```diff
@@ -61,19 +61,43 @@ func (a *Blocks) HeightBySignature(ctx context.Context, id string) (*BlocksHeigh
...
-	NxtConsensus     NxtConsensus     `json:"nxt-consensus"`
-	Features         []uint64         `json:"features"`
-	Generator        proto.Address    `json:"generator"`
-	Signature        crypto.Signature `json:"signature"`
-	Blocksize        uint64           `json:"blocksize"`
-	TransactionCount uint64           `json:"transactionCount"`
-	Height           uint64           `json:"height"`
-
-	ID proto.BlockID `json:"id"`
+	Version            uint64           `json:"version"`
```
alexeykiselev, line 84: Shouldn't we put `omitempty` on new fields here in order to support compatibility with previous version of block.

PR #358, opened 2020-09-21
added groth16Verify on the bls12-381 and the bn256 curve

esuwu:


pkg/crypto/internal/groth16/bls12381/groth16Verify.go
```diff
@@ -0,0 +1,81 @@
+package bls12381
+
+import (
+	"errors"
+	"github.com/consensys/gurvy/bls381/fr"
+	Proof "github.com/wavesplatform/gowaves/pkg/crypto/internal/groth16/bls12381/proof"
```
alexeykiselev, line 6: Quote: "When renaming an imported package, the local name should follow the same guidelines as package names (lower case, no under_scores or mixedCaps)" from https://blog.golang.org/package-names.
Why rename if there is no package name collision? If it collides with variable names, like proof in this case, better rename variable to something short, e.g. `proof` to `p`.

pkg/crypto/internal/groth16/bls12381/proof/proof.go
```diff
@@ -0,0 +1,55 @@
+package proof
```
alexeykiselev, line 1: I think it's better no to separate `proof` and `verificationKey`(wrong naming also), but join them in one package `internal/groth16/bls12-381/` for example, then use renaming to not collide with `kilic/bls12-381`.

pkg/crypto/internal/groth16/bls12381/verificationKey/verificationKey.go
```diff
@@ -0,0 +1,90 @@
+package verificationKey
```
alexeykiselev, line 1: Move to one package

pkg/crypto/internal/groth16/bls12381/verifier/verifier.go
```diff
@@ -0,0 +1,30 @@
+package verifier
```
alexeykiselev, line 1: Move

alexeykiselev (changes requested):

esuwu:
Benchamarks для 0, 1, 15, 16 inputs. Запускался на i7
Снимок экрана 2020-09-22 в 16 15 54

alexeykiselev (approved):

PR #380, opened 2020-10-26
Added correct range for GetBlockRange

esuwu:


esuwu:
I was afraid about to write 
```
stream, err := s.blockRange(h, h+delta-1, ctx, true)  - h + delta  - 1
```
But I saw correct behavior by the end of loop, so it seems like it's okay - we will receive the last block on the height == end.
Screenshot from 2020-10-26 20-02-50

alexeykiselev (approved):

PR #387, opened 2020-11-09
Ride invoke dApp from dApp

esuwu:


pkg/ride/tree_evaluation.go
```diff
@@ -13,6 +13,32 @@ func CallVerifier(env RideEnvironment, tree *Tree) (RideResult, error) {
 	return e.evaluate()
 }
 
+func invokeFunctionFromDApp(env RideEnvironment, recipient proto.Recipient, fnName rideType, args []rideType) (RideResult, error) {
+	var funcName string
```
alexeykiselev, line 17:
```
funcName, ok := fnName.(rideString)
if !ok {
  return nil, errors...
}
```
Isn't that shorter? And convert `funcName` to string later on use.

pkg/state/state.go
```diff
@@ -464,6 +464,18 @@ func newStateManager(dataDir string, params StateParams, settings *settings.Bloc
 	return state, nil
 }
 
+func (s *stateManager) GetByteTree(recipient proto.Recipient) (proto.Script, error) {
+	if recipient.Address != nil {
```
alexeykiselev, line 468: Better not just check that Recipient is an address, but lookup for an address here by alias from Recipient.

pkg/ride/tree_evaluator.go
```diff
@@ -209,11 +209,19 @@ type treeEvaluator struct {
 	env RideEnvironment
 }
 
+var wasInvokeDAppFromDApp = false
```
alexeykiselev, line 212: Don't get this. What's the difference for function invoked from dapp or not.

pkg/ride/tree_evaluator.go
```diff
@@ -384,6 +388,36 @@ func (e *treeEvaluator) walk(node Node) (rideType, error) {
...
+					return a, nil
+				}
+				args[i] = a
+			}
+
+			res, err := invokeFunctionFromDApp(e.env, recipient, funcName, args[2:])
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to get RideResult from invokeFunctionFromDApp of system function '%s'", id)
+			}
+			resultDAppFromDApp = res
```
alexeykiselev, line 416: As I told you, check `res` here for an error and add ScriptAcitons to evaluator's actions list in case of no error.

pkg/types/types.go
```diff
@@ -33,12 +33,17 @@ type TransactionWithBytes struct {
 	B []byte
 }
 
+type Node interface {
```
alexeykiselev, line 36: How this interface is used?

pkg/ride/environment.go
```diff
@@ -140,7 +325,225 @@ func (e *Environment) txID() rideType {
...
+				e.st.diff.dataEntries.diffString = append(e.st.diff.dataEntries.diffString, stringEntry)
+			}
+
+			if res.Entry.GetValueType() == proto.DataDelete {
+
+				key := res.Entry.GetKey()
+
+				for i, intDataEntry := range e.st.diff.dataEntries.diffInteger {
+					if key == intDataEntry.Key {
+						length := len(e.st.diff.dataEntries.diffInteger)
```
alexeykiselev, line 478: No, no, no! Deleting doesn't mean to delete entry from local storage, but replace it with empty value. Removing from local storage just lead to retrieving it from state.

pkg/ride/environment.go
```diff
@@ -140,7 +325,225 @@ func (e *Environment) txID() rideType {
...
+						e.st.diff.dataEntries.diffBinary = e.st.diff.dataEntries.diffBinary[:length-1]
+
+						return nil
+					}
+				}
+
+			}
+		case proto.TransferScriptAction:
+
+			if searchBalance := e.st.diff.findBalance(res.Recipient, res.Asset.ID.Bytes()); searchBalance != nil {
```
alexeykiselev, line 523: Why for we need this `searchBalance`? It's not used.

pkg/ride/environment.go
```diff
@@ -140,7 +325,225 @@ func (e *Environment) txID() rideType {
...
+			} else {
+				var balance diffBalance
+
+				balance.recipient = res.Recipient
+				balance.assetID = res.Asset.ID
+				balance.amount = res.Amount
+				e.st.diff.balances = append(e.st.diff.balances, balance)
+			}
+
+			if searchWavesBalance := e.st.diff.findWavesBalance(res.Recipient); searchWavesBalance != nil {
```
alexeykiselev, line 534: The same, why update if it's not used.

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #394, opened 2020-12-01
Fix incorrect data on stagenet

alexeykiselev:


pkg/ride/functions_proto.go
```diff
@@ -381,12 +381,17 @@ func addressToString(_ RideEnvironment, args ...rideType) (rideType, error) {
 	if err := checkArgs(args, 1); err != nil {
 		return nil, errors.Wrap(err, "addressToString")
 	}
-	a, ok := args[0].(rideAddress)
-	if !ok {
+	switch a := args[0].(type) {
+	case rideAddress:
+		return rideString(proto.Address(a).String()), nil
+	case rideRecipient:
+		if a.Address == nil {
```
esuwu, line 388: There could be Alias, probably you should check it especially when a.Address == nil

alexeykiselev: Nope, the function accepts only an address to convert it into string, so Alias is not expected here.

Frozen (approved):

PR #430, opened 2021-02-25
Fix microblock signature

alexeykiselev:


pkg/proto/block_test.go
```diff
@@ -362,14 +362,13 @@ func TestBlock_Clone(t *testing.T) {
...
 func TestEmptyBlockMarshall(t *testing.T) {
-	t.Skip()
 	b1 := Block{}
 	bts, err := b1.MarshalBinary()
 	require.NoError(t, err)
 
 	b2 := Block{}
 	err = b2.UnmarshalBinary(bts, MainNetScheme)
-	require.NoError(t, err)
+	require.Error(t, err)
```
Frozen, line 371: We marshal successfully above, but can't unmarshal later, it's strange behavior.

pkg/metrics/metrics.go
```diff
@@ -332,6 +547,15 @@ func reportBlock(t tags, f fields) {
 	rep.in <- p
 }
 
+func reportFSM(t tags, f fields) {
+	p, err := influx.NewPoint("fsm", t, f, time.Now())
+	if err != nil {
+		zap.S().Warn("Failed to create metrics point 'fsm': %v", err)
+		return
+	}
+	rep.in <- p
```
Frozen, line 556: if `rep.in` is full, it can hang forever.

Frozen (approved):

pkg/proto/block_test.go
alexeykiselev: It's written in TODO

Frozen (approved):

PR #433, opened 2021-03-04
Added error loggs handling

esuwu:


cmd/forkdetector/internal/api.go
```diff
@@ -107,7 +107,7 @@ func (a *api) Start() <-chan struct{} {
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 		err := a.srv.Shutdown(ctx)
 		if err != nil {
-			zap.S().Errorf("Failed to shutdown API server: %v", err)
+			zap.S().Infof("Failed to shutdown API server: %v", err)
```
alexeykiselev, line 110: No need to change it here.

cmd/wmd/internal/api.go
```diff
@@ -99,7 +99,7 @@ func NewDataFeedAPI(interrupt <-chan struct{}, logger *zap.Logger, storage *stat
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 		err := apiServer.Shutdown(ctx)
 		if err != nil {
-			zap.S().Errorf("Failed to shutdown API server: %v", err)
+			zap.S().Infof("Failed to shutdown API server: %v", err)
```
alexeykiselev, line 102: No need to change it here

pkg/api/node_api.go
```diff
@@ -196,7 +196,7 @@ func Run(ctx context.Context, address string, n *NodeApi) error {
 		zap.S().Info("Shutting down API...")
 		err := apiServer.Shutdown(ctx)
 		if err != nil {
-			zap.S().Errorf("Failed to shutdown API server: %v", err)
+			zap.S().Infof("Failed to shutdown API server: %v", err)
```
alexeykiselev, line 199: Better to put it on WARN level

pkg/node/node.go
```diff
@@ -100,6 +100,15 @@ func (a *Node) Serve(ctx context.Context) error {
 	}
 }
 
+func (a *Node) LogErrors(err error) {
```
alexeykiselev, line 103: Definitely this is not a public function. May be there is no need for a separate function at all if it used once...

alexeykiselev (changes requested):

pkg/node/node.go
esuwu: You're right but the function "Run" is big enough I'll make that method private

alexeykiselev (approved):

PR #438, opened 2021-03-18
Ride vm and compiler.

Frozen:


Makefile
```diff
@@ -11,7 +11,7 @@ export GO111MODULE=on
 
 .PHONY: vendor vetcheck fmtcheck clean build gotest
 
-all: vendor vetcheck fmtcheck build gotest mod-clean
```
esuwu, line 14: It's better to skip a package instead of removing fmtcheck at all in case if it's hard to fix style and stuff

Frozen: I've tried all possible ways - skip, ignore, fix, upgrade. Nothing helps.

PR #452, opened 2021-04-12
Deleted extra fee for smart asset script since RideV5

esuwu:


pkg/state/fee_validation.go
```diff
@@ -144,6 +144,10 @@ func newTxCosts(smartAssets, smartAccounts uint64, isRideV5Activated bool, compl
 	smartAssetsFee := smartAssets * scriptExtraFee
 	smartAccountsFee := smartAccounts * scriptExtraFee
 
+	if isRideV5Activated {
+		smartAssetsFee = smartAssets // since RideV5 we have to erase extra fee for smart asset scripts
```
alexeykiselev, line 148: Shouldn't we set `smartAssetsFee` to 0 here?

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #459, opened 2021-04-22
Additional RIDE functions to retrieve data entries on this account imp…

alexeykiselev:
…lemented and tested


pkg/ride/functions_proto.go
```diff
@@ -310,6 +310,23 @@ func intFromState(env Environment, args ...rideType) (rideType, error) {
...
 }
 
+func intFromSelfState(env Environment, args ...rideType) (rideType, error) {
+	k, err := extractKey(args)
+	if err != nil {
+		return rideUnit{}, nil
+	}
+	a, ok := env.this().(rideAddress)
+	if !ok {
+		return rideUnit{}, nil
```
esuwu, line 320: let's return a specifc error like "failed to convert argument to ride address" or something like this

pkg/ride/functions_proto.go
```diff
@@ -310,6 +310,23 @@ func intFromState(env Environment, args ...rideType) (rideType, error) {
...
+	}
+	r := proto.NewRecipientFromAddress(proto.Address(a))
+	entry, err := env.state().RetrieveNewestIntegerEntry(r, k)
+	if err != nil {
+		return rideUnit{}, nil
```
esuwu, line 325: return err or errors.Wrap(err, ...)

pkg/ride/functions_proto.go
```diff
@@ -310,6 +310,23 @@ func intFromState(env Environment, args ...rideType) (rideType, error) {
 	return rideInt(entry.Value), nil
 }
 
+func intFromSelfState(env Environment, args ...rideType) (rideType, error) {
+	k, err := extractKey(args)
+	if err != nil {
+		return rideUnit{}, nil
```
esuwu, line 316: return err or errors.Wrap(err, ...

pkg/ride/functions_proto.go
```diff
@@ -322,6 +339,23 @@ func bytesFromState(env Environment, args ...rideType) (rideType, error) {
 	return rideBytes(entry.Value), nil
 }
 
+func bytesFromSelfState(env Environment, args ...rideType) (rideType, error) {
+	k, err := extractKey(args)
+	if err != nil {
+		return rideUnit{}, nil
```
esuwu, line 345: why did you decided to not return error we could get?

esuwu (changes requested):

esuwu:
there are many cases where we probably should return an error

pkg/ride/generate/main.go
```diff
@@ -488,13 +488,21 @@ func functionsV5() map[string]string {
...
+	m["1056"] = "booleanFromSelfState"
+	m["1057"] = "bytesFromSelfState"
+	m["1058"] = "stringFromSelfState"
 	m["1081"] = "calculateLeaseID"
 	m["1092"] = "simplifiedLease"
 	m["1093"] = "fullLease"
 	m["LeaseCancel"] = "leaseCancel"
-	m["1054"] = "isDataStorageUntouched"
-	m["1009"] = "hashScriptAtAddress"
+	m["@extrNative(1055)"] = "intValueFromSelfState"
```
esuwu, line 502: why do we need both "intValueFromSelfState" and "intFromSelfState"? Don't they do literally the same?

pkg/ride/functions_proto.go
alexeykiselev: No, it's a special class of functions that swallows all errors.

pkg/ride/functions_proto.go
alexeykiselev: Same reason, it have to return `Unit` on every error.

pkg/ride/functions_proto.go
alexeykiselev: Nope

pkg/ride/functions_proto.go
alexeykiselev: Because this functions work like this in Scala node.

pkg/ride/generate/main.go
alexeykiselev: Because the first one will return `throw` on `Unit` value, but the latter won't.

esuwu (approved):

PR #462, opened 2021-05-14
Node http api improvements

nickeskov:


pkg/api/app.go
```diff
@@ -97,28 +97,32 @@ func (a *App) LoadKeys(apiKey string, password []byte) error {
...
 func (a *App) checkAuth(key string) error {
+	// TODO(nickeskov): use new types of errors
 	if !a.apiKeyEnabled {
 		return &AuthError{errors.New("api key disabled")}
 	}
 	d, err := crypto.SecureHash([]byte(key))
 	if err != nil {
-		return &AuthError{err}
+		// TODO(nickeskov): it's OK?
+		return err
```
esuwu, line 125: you wanted to use custom errors in order to make switch on them. Why is here just a regular error?

pkg/api/app_miner.go
```diff
@@ -19,18 +19,18 @@ type MinerInfo struct {
 	Scheduler Scheduler
 }
 
-func (a *App) Miner() (*MinerInfo, error) {
+func (a *App) Miner() (MinerInfo, error) {
```
esuwu, line 22: this function is not going to return an error. There is always `nil`. Should be `func (a *App) Miner() MinerInfo`

pkg/api/app_peers.go
```diff
@@ -14,22 +14,22 @@ type Peer struct {
...
+type PeersKnown struct {
 	Peers []Peer `json:"peers"`
 }
 
-func (a *App) PeersAll() (*PeersAll, error) {
+func (a *App) PeersKnown() (PeersKnown, error) {
 	peers, err := a.state.Peers()
 	if err != nil {
-		return nil, &InternalError{err}
+		return PeersKnown{}, errors.Wrap(err, "PeersKnown")
```
esuwu, line 24: the returned error should be errors.Wrap(err, "failed to get known peers")

pkg/api/app_peers.go
```diff
@@ -73,24 +73,29 @@ type PeersConnectedRow struct {
 	ApplicationVersion string `json:"applicationVersion"`
 }
 
-func (a *App) PeersConnected() (*PeersConnectedResponse, error) {
+func (a *App) PeersConnected() (PeersConnectedResponse, error) {
```
esuwu, line 76: this function is also not going to return an error

pkg/api/app_peers.go
```diff
@@ -99,16 +104,16 @@ type PeersSuspendedResponse struct {
 	Peers []string `json:"peers"`
 }
 
-func (a *App) PeersSuspended() (*PeersSuspendedResponse, error) {
+func (a *App) PeersSuspended() (PeersSuspendedResponse, error) {
```
esuwu, line 107: the same about errors

esuwu: are you sure there should be a particular function for calling `a.peers.Suspended()`?

pkg/api/errors.go
```diff
@@ -11,3 +21,69 @@ type AuthError struct {
...
+func (eh *ErrorHandler) Handle(w http.ResponseWriter, r *http.Request, err error) {
+	if err == nil {
+		return
+	}
+	switch innerErr := errors.Cause(err).(type) {
+	case BadRequestError, *BadRequestError:
+		// TODO(nickeskov): remove this
+		http.Error(w, fmt.Sprintf("Failed to complete request: %s", innerErr.Error()), http.StatusForbidden)
+	case AuthError, *AuthError:
+		// TODO(nickeskov): remove this
```
esuwu, line 44: `// TODO(nickeskov): remove this` when exactly are you going to remove `this`? delete comments here or remove something that you was going to

pkg/api/errors.go
```diff
@@ -11,3 +21,69 @@ type AuthError struct {
...
+	w.WriteHeader(apiErr.GetHttpCode())
+	if encodeErr := json.NewEncoder(w).Encode(apiErr); encodeErr != nil {
+		eh.logger.Error("Failed to marshal API Error to JSON",
+			zap.String("proto", r.Proto),
+			zap.String("path", r.URL.Path),
+			zap.String("request_id", middleware.GetReqID(r.Context())),
+			zap.Error(encodeErr),
+			zap.String("api_error", apiErr.Error()),
+		)
+		panic(errors.Errorf("BUG, CREATE REPORT: %s", encodeErr.Error()))
```
esuwu, line 87: should we really panic here?

pkg/api/errors/transaction.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"github.com/wavesplatform/gowaves/pkg/proto"
+	"net/http"
+)
+
+type transactionError struct {
+	genericError
+}
+
+// TODO(nickeskov): IMPLEMENT ME
```
esuwu, line 12: should it be implemented if it's not used?

pkg/api/app_blocks.go
```diff
@@ -9,12 +10,12 @@ type Score struct {
 	Score string `json:"score"`
 }
 
-func (a *App) BlocksScoreAt(at proto.Height) (*Score, error) {
+func (a *App) BlocksScoreAt(at proto.Height) (Score, error) {
```
esuwu, line 13: I don't think here should be a function for just `score, err := a.state.ScoreAtHeight(at)`

pkg/api/errors.go
nickeskov: Yes, we should. Type which implements `ApiError` interface must be serializable to JSON. If encode fails that is definitely a bug.

pkg/api/app.go
nickeskov: Regular errors will be automatically transformed into custom unknown error, so here I want to use regular error.

pkg/api/app_blocks.go
nickeskov: I don't think so because this layer hides node and state internals.

pkg/api/errors/transaction.go
nickeskov: Yes because we need to support scala node HTTP API.

pkg/api/app_peers.go
alexeykiselev: Better to change error message here

alexeykiselev (changes requested):
Please, rebase on `version-0.9` branch

nickeskov:
See #475

PR #486, opened 2021-06-07
Cancel leases to stolen aliases

alexeykiselev:


nickeskov (changes requested):

pkg/state/keys.go
```diff
@@ -390,6 +390,21 @@ func (k *disabledAliasKey) bytes() []byte {
...
 
+func (k *disabledAliasKey) unmarshal(data []byte) error {
+	if len(data) != disabledAliasKeySize {
+		return errInvalidDataSize
+	}
+	if data[0] != disabledAliasKeyPrefix {
+		return errInvalidPrefix
+	}
+	var err error
+	k.alias, err = proto.StringWithUInt16Len(data[1:])
```
esuwu, line 401: ```k.alias, err := proto.StringWithUInt16Len(data[1:])```

pkg/state/leases.go
```diff
@@ -37,38 +41,19 @@ func (lr *leaseRecordForStateHashes) less(other stateComponent) bool {
...
-	l.isActive, err = proto.Bool(data[0:1])
-	if err != nil {
-		return err
-	}
-	l.leaseAmount = binary.BigEndian.Uint64(data[1:9])
-	copy(l.recipient[:], data[9:9+proto.AddressSize])
-	copy(l.sender[:], data[9+proto.AddressSize:9+proto.AddressSize*2])
-	return nil
+func (l leasing) isActive() bool {
+	return l.Status == LeaseActive
```
esuwu, line 56: this construction is weird. Shouldn't it be ``` return l.Status``` ?

pkg/state/keys.go
alexeykiselev: Won't work here, `err` have to be declared beforehand because only assignment can be used to assign value to `k.alias`.

pkg/state/leases.go
alexeykiselev: Nope, one status is active, the other two is inactive. For now we only want to know is it active or not. May be we will have to remove third status and move back to simple bool value. Don't know.

nickeskov (approved):

PR #494, opened 2021-06-21
Created ethereum transaction and supported translation of Ride DApp metadata to ethereum ABI. Created metamask service stub.

esuwu:


pkg/metamask/server.go
```diff
@@ -0,0 +1,35 @@
...
+	rpc := zenrpc.NewServer(zenrpc.Options{ExposeSMD: true})
+	rpc.Register("", MetaMask{}) // public
+	rpc.Use(zenrpc.Logger(log.New(os.Stderr, "", log.LstdFlags)))
+
+	http.Handle("/", rpc)
+
+	server := &http.Server{Addr: address, Handler: nil}
+
+	go func() {
+		zap.S().Info("shutting down metamask service...")
```
nickeskov, line 23: You should wait for `ctx.Done`. Or you should write this line after `err := server.Shutdown(ctx)`.

pkg/metamask/server.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func RunMetaMaskService(ctx context.Context, address string) error {
+	rpc := zenrpc.NewServer(zenrpc.Options{ExposeSMD: true})
+	rpc.Register("", MetaMask{}) // public
+	rpc.Use(zenrpc.Logger(log.New(os.Stderr, "", log.LstdFlags)))
+
+	http.Handle("/", rpc)
+
+	server := &http.Server{Addr: address, Handler: nil}
```
nickeskov, line 20: You should handle HTTP OPTIONS method.

nickeskov (changes requested):

pkg/metamask/server.go
```diff
@@ -0,0 +1,36 @@
...
+	rpc := zenrpc.NewServer(zenrpc.Options{ExposeSMD: true, AllowCORS: true})
+	rpc.Register("", MetaMask{}) // public
+	rpc.Use(zenrpc.Logger(log.New(os.Stderr, "", log.LstdFlags)))
+
+	http.Handle("/", rpc)
+
+	server := &http.Server{Addr: address, Handler: nil}
+
+	go func() {
+		err := server.Shutdown(ctx)
```
nickeskov, line 23: You should wait for `ctx.Done()`. For example:
```go
<-ctx.Done()
zap.S().Info("shutting down metamask service...")
err := server.Shutdown(ctx)
// ....
```

nickeskov (changes requested):

cmd/node/node.go
```diff
@@ -59,6 +60,7 @@ var (
 	apiKey                                = flag.String("api-key", "", "Api key")
 	grpcAddr                              = flag.String("grpc-address", "127.0.0.1:7475", "Address for gRPC API")
 	enableGrpcApi                         = flag.Bool("enable-grpc-api", true, "Enables/disables gRPC API")
+	enableMetaMaskService                 = flag.Bool("enable-metamask", true, "Enables/disables metamask service")
```
nickeskov, line 63: I think `metaMaskServiceAddr` parameter would be useful (Address for metamask service).

cmd/node/node.go
```diff
@@ -417,6 +419,17 @@ func main() {
 		}()
 	}
 
+	address := ":8245"
```
nickeskov, line 422: Use 8545 port as default.

nickeskov (changes requested):

pkg/metamask/server.go
```diff
@@ -0,0 +1,36 @@
...
+	"go.uber.org/zap"
+	"log"
+	"net/http"
+	"os"
+)
+
+func RunMetaMaskService(ctx context.Context, address string) error {
+	rpc := zenrpc.NewServer(zenrpc.Options{ExposeSMD: true, AllowCORS: true})
+	rpc.Register("", MetaMask{}) // public
+	rpc.Use(zenrpc.Logger(log.New(os.Stderr, "", log.LstdFlags)))
```
nickeskov, line 16: Maybe use zap logger?

pkg/metamask/server.go
esuwu: I shouldn't. <-ctx.Done() is already in body of function server.Shutdown(ctx)

nickeskov: Nope, you should. In first lines in body of function `server.Shutdown(ctx)` you can see that server close listeners. It means that new incoming requests will be rejected.

pkg/metamask/abi/parsing_test.go
```diff
@@ -40,7 +41,25 @@ func TestJsonAbi(t *testing.T) {
...
 	require.NoError(t, err)
 
-	resJson, err := getJsonAbi(callData.Signature)
+	resJson, err := getJsonAbi(callData.Signature, callData.Inputs)
+	require.NoError(t, err)
+	require.Equal(t, string(resJson), expectedJson)
+}
+
+func TestJsonAbiPayments(t *testing.T) {
+	expectedJson := `[{"name":"transfer","type":"function","inputs":[{"type":"address"},{"type":"uint256"},{"type":"[(address, uint256)]"}]}]`
```
nickeskov, line 50: Payments ABI-JSON representation isn't correct. See [this](https://github.com/ethereum/eth-abi/issues/35) for example.

nickeskov (approved):

PR #496, opened 2021-06-23
Address structure renamed to WavesAddress. New interface Address intr…

alexeykiselev:
…oduced. New type AddressID added and used to identify Address in state.


pkg/proto/address.go
```diff
@@ -17,179 +18,221 @@ import (
...
+	return b[:]
+}
+
+func (b EthereumAddress) ID() AddressID {
+	var id AddressID
+	copy(id[:], b[:])
+	return id
+}
+
+func (b EthereumAddress) String() string {
```
nickeskov, line 70: Stringer implementation for EthereumAddress isn't correct. See correct implementation in go-ethereum repository. (But maybe for our purposes we don't need correct implementation of this method.)

nickeskov (dismissed):

nickeskov:
See #561

PR #497, opened 2021-06-23
Fix ride version check

alexeykiselev:


pkg/proto/transactions_with_proofs.go
```diff
@@ -209,7 +211,7 @@ func (tx *IssueWithProofs) Validate() (Transaction, error) {
 			if !validContentType(tx.Script[1]) {
 				return tx, errors.Errorf("Invalid content type of script: %d", tx.Script[1])
 			}
-			if tx.Script[2] > 4 { // 4 is current max script version
+			if tx.Script[2] > topRideVersion {
```
nickeskov, line 214: Maybe use `validStdVersion` function?

nickeskov (approved):

pkg/proto/transactions_with_proofs.go
alexeykiselev: That's a good point

nickeskov (approved):

PR #512, opened 2021-07-13
Limit new outbound connections

alexeykiselev:


pkg/util/fdlimit/fdlimit_unix.go
```diff
@@ -17,6 +17,9 @@ func SetMaxFDs(max uint64) (uint64, error) {
 	if max > rLimit.Max {
 		max = rLimit.Max
 	}
+	if rLimit.Cur > max {
+		return rLimit.Cur, nil
+	}
```
nickeskov, line 22: I think that isn't correct. In this case caller can't decrease max file descriptors amount.

pkg/util/fdlimit/fdlimit_darwin.go
```diff
@@ -0,0 +1,50 @@
...
+		return 0, errors.Wrap(err, "error getting rlimit")
+	}
+
+	// Try to update the rLimit to the max allowance
+	if max > darwinMaxRLimit {
+		max = darwinMaxRLimit
+	}
+	if rLimit.Cur > max {
+		return rLimit.Cur, nil
+	}
```
nickeskov, line 24: This is the same issue that you can found in `fdlimit_unix.go` file.

pkg/util/fdlimit/fdlimit_darwin.go
```diff
@@ -0,0 +1,50 @@
...
+package fdlimit
+
+import (
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const darwinMaxRLimit = 4096
+
+func SetMaxFDs(max uint64) (uint64, error) {
```
nickeskov, line 12: On my mind RaiseMaxFDs would be a better name for this function than SetMaxFDs.

nickeskov (dismissed):

nickeskov (dismissed):

nickeskov (approved):

PR #540, opened 2021-08-27
OrderV4 with 'Eip712Signature' protobuf field support.

nickeskov:


pkg/ride/converters.go
```diff
@@ -397,14 +397,23 @@ func orderToObject(scheme proto.Scheme, o proto.Order) (rideObject, error) {
...
-	body, err := proto.MarshalOrderBody(scheme, o)
+	// note that in ride we use only proto.WavesAddress addresses
+	senderWavesAddr, err := senderAddr.ToWavesAddress(scheme)
 	if err != nil {
-		return nil, errors.Wrap(err, "orderToObject")
+		return nil, errors.Wrapf(err, "failed to transform (%T) address type to WavesAddress type", senderAddr)
+	}
+	var body []byte
+	if _, ok := o.(*proto.EthereumOrderV4); ok {
+		body = nil
```
esuwu, line 411: there is no much sense in `body = nil`. It's nil by default. 
```
if _, ok := o.(*proto.EthereumOrderV4); !ok {
        body, err = proto.MarshalOrderBody(scheme, o)
        if err != nil {
	   return nil, errors.Wrap(err, "orderToObject")
        }
```

pkg/ride/converters.go
```diff
@@ -1347,13 +1356,20 @@ func signatureToProofs(sig *crypto.Signature) RideList {
 
 func proofs(proofs *proto.ProofsV1) RideList {
 	r := make(RideList, 8)
```
esuwu, line 1358: if you don't to `make(RideList, 8)` for just 8 cells, there will be no need to do this
```
else {
		// special case for ethereum stuff
		for i := range r {
			r[i] = RideBytes(nil)
		}
	}
```

esuwu: Is there any specific point to do `every r[i] = RideBytes(nil)` instead of just `r = nil`?

pkg/proto/eth_math.go
```diff
@@ -0,0 +1,104 @@
...
+
+// parseEthereumBig256 parses s as a 256 bit integer in decimal or hexadecimal syntax.
+// Leading zeros are accepted. The empty string parses as zero.
+func parseEthereumBig256(s string) (*big.Int, bool) {
+	if s == "" {
+		return new(big.Int), true
+	}
+	var bigint *big.Int
+	var ok bool
+	if len(s) >= 2 && (s[:2] == "0x" || s[:2] == "0X") {
```
esuwu, line 45: the function you're looking for is `strings.TrimPrefix(s, "0x")`
instead of `(s[:2] == "0x" || s[:2] == "0X")` and `bigint, ok = new(big.Int).SetString(s[2:], 16)`

pkg/ride/converters.go
nickeskov: In short, yes. We should set empty value for each element of `RideList` according to the specification.

pkg/proto/eth_math.go
nickeskov: The main point is detect and trim `0x` or `0X` prefix. In your version I can't detect and trim this prefixes in the short and convenient way. I have to implement both variants of `strings.TrimPrefix(s, "0x")` and `strings.TrimPrefix(s, "0X")`.

esuwu (approved):

alexeykiselev (approved):

alexeykiselev (approved):

PR #547, opened 2021-10-14
New implementations with new math contexts for Pow and PowBigInt func…

alexeykiselev:
…tions. New test on corner cases. New benchmarks of Pow function.

Old implementation of Pow function preserved as PowV1 for backward compatibility. Test on backward compatibility added.

### DIFF
--- a/pkg/ride/math/math.go
+++ b/pkg/ride/math/math.go
@@ -1,6 +1,7 @@
 package math
 
 import (
+	math0 "math"
 	"math/big"
 
 	"github.com/ericlagergren/decimal"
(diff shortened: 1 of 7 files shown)


pkg/ride/math/math.go
```diff
@@ -43,53 +112,84 @@ func convertToBigIntResult(v *decimal.Big, scale int, mode decimal.RoundingMode)
...
+	if !checkScales(baseScale, exponentScale, resultScale) {
+		return 0, errors.New("invalid scale")
+	}
+	b := decimal.WithContext(longContext).SetMantScale(base, baseScale)
+	e := decimal.WithContext(longContext).SetMantScale(exponent, exponentScale)
+	r, err := pow(b, e, longContext)
+	if err != nil {
+		return 0, err
+	}
+	r, err = rescale(r, resultScale, 19, longContext)
```
nickeskov, line 157: I think using of `longContext.Precision` would be better than `19`.

pkg/ride/math/math.go
```diff
@@ -43,53 +112,84 @@ func convertToBigIntResult(v *decimal.Big, scale int, mode decimal.RoundingMode)
...
-	e := decimal.WithContext(decimal.Context128).SetBigMantScale(exponent, exponentScale)
-	r, err := pow(b, e)
+	b := decimal.WithContext(bigIntContext).SetBigMantScale(base, baseScale)
+	e := decimal.WithContext(bigIntContext).SetBigMantScale(exponent, exponentScale)
+	r, err := pow(b, e, bigIntContext)
 	if err != nil {
 		return nil, err
 	}
-	return convertToBigIntResult(r, resultScale, mode)
+	r, err = rescale(r, resultScale, 154, bigIntContext)
```
nickeskov, line 181: I think using of `bigIntContext.Precision` would be better than `154`.

pkg/ride/math/math.go
```diff
@@ -43,53 +112,84 @@ func convertToBigIntResult(v *decimal.Big, scale int, mode decimal.RoundingMode)
...
+		return nil, err
+	}
+	if !r.IsNormal() {
+		return nil, errors.New("not normal")
+	}
+	context := bigIntContext
+	context.RoundingMode = mode
+	r = decimal.WithContext(context).Set(r)
+	return r.RoundToInt().Int(nil), nil
+	//return convertToBigIntResult(r, resultScale, mode)
```
nickeskov, line 192: Unnecessary comment.

PR #564, opened 2021-11-03
Added script basic info entity to state

nickeskov:


pkg/state/scripts_storage.go
```diff
@@ -69,40 +69,56 @@ func (as *assetScripRecordForHashes) less(other stateComponent) bool {
...
 
-func scriptExists(recordBytes []byte) bool {
-	// Detect if script length is not 0 without unmarshal.
-	return len(recordBytes) > crypto.KeySize
-}
-
 type scriptRecord struct {
-	pk     crypto.PublicKey
-	script proto.Script
+	Data proto.Script `cbor:"0,keyasint,omitemtpy"`
```
alexeykiselev, line 73: I think it simpler to store script bytes without wrapping them in CBOR. Do we plan to extend this structure later? Because it looks like this is the only reason to use CBOR encoding here. Giving up CBOR here will save us a little time on encoding/decoding.

alexeykiselev: Consider to remove `scriptRecord` completely.

pkg/state/scripts_storage.go
```diff
@@ -69,40 +69,56 @@ func (as *assetScripRecordForHashes) less(other stateComponent) bool {
...
+func (r *scriptBasicInfoRecord) marshalBinary() ([]byte, error) {
+	return cbor.Marshal(r)
+}
+
+func (r *scriptBasicInfoRecord) unmarshalBinary(data []byte) error {
+	return cbor.Unmarshal(data, r)
+}
+
+type scriptDBItem struct {
+	script scriptRecord
```
alexeykiselev, line 113: Replace `scriptRecord` here with just `[]byte` or `Script`

pkg/state/scripts_storage.go
```diff
@@ -131,28 +147,37 @@ func newScriptsStorage(hs *historyStorage, calcHashes bool) (*scriptsStorage, er
...
+	scriptKeyBytes := key.bytes()
+	if err := ss.hs.addNewEntry(scriptType, scriptKeyBytes, scriptRecordBytes, blockID); err != nil {
 		return err
 	}
-	if record.scriptIsEmpty() {
+	scriptBasicInfoKeyBytes := (&scriptBasicInfoKey{scriptKey: key}).bytes()
+	if err := ss.hs.addNewEntry(scriptBasicInfo, scriptBasicInfoKeyBytes, scriptBasicInfoRecordBytes, blockID); err != nil {
+		return err
+	}
+	if dbItem.script.isEmpty() {
```
alexeykiselev, line 171: `len(dbItem.script) == 0` will be sufficient here

pkg/state/scripts_storage.go
```diff
@@ -347,24 +375,32 @@ func (ss *scriptsStorage) accountHasVerifier(addr proto.Address, filter bool) (b
...
+	}
+	return info.scriptExists(), nil
 }
 
 func (ss *scriptsStorage) accountHasScript(addr proto.Address, filter bool) (bool, error) {
-	key := accountScriptKey{addr}
+	key := scriptBasicInfoKey{scriptKey: &accountScriptKey{addr}}
 	recordBytes, err := ss.hs.topEntryData(key.bytes(), filter)
-	if err != nil {
+	if err != nil { // TODO: check error type
```
alexeykiselev, line 396: Do we left something to do?

alexeykiselev (approved):

PR #584, opened 2021-11-12
Ride feature 17 check

esuwu:


nickeskov:
This branch was used only for tests.

PR #587, opened 2021-11-16
Changed receivers of 'GetID' method of 'Transaction' interface.

nickeskov:


pkg/proto/transactions.go
```diff
@@ -458,7 +458,7 @@ func (tx *Genesis) generateID(scheme Scheme) error {
 	return nil
 }
 
-func (tx Genesis) GetID(scheme Scheme) ([]byte, error) {
+func (tx *Genesis) GetID(scheme Scheme) ([]byte, error) {
```
alexeykiselev, line 461: Is it OK to have a mixture of different receivers? May be better to make all of them pointers?

alexeykiselev (commented):
Do we really need to have a non-pointer receivers for transactions?

pkg/proto/transactions.go
nickeskov: I'll do it in new PR.

alexeykiselev (approved):

PR #592, opened 2021-11-23
Ride V6 activation and code generation added.

alexeykiselev:
NativeFold implemented. Simple tests on native fold added.


pkg/ride/tree_estimatorV4.go
```diff
@@ -0,0 +1,260 @@
...
+	case "452":
+		return 100, 2, true
+	case "453":
+		return 200, 2, true
+	case "454":
+		return 500, 2, true
+	case "455":
+		return 1000, 2, true
+	default:
+		return 0, 0, false
```
nickeskov, line 258: Add information (comments) about functions IDs.

pkg/ride/tree_evaluator.go
```diff
@@ -285,40 +399,86 @@ func (e *treeEvaluator) evaluateNativeFunction(name string, arguments []Node) (r
...
+	}
+	if len(args) != len(uf.Arguments) {
+		return nil, errors.Errorf("mismatched arguments number of user function '%s'", name)
+	}
+	avs := make([]esValue, len(args))
+	for i, arg := range args {
+		an := uf.Arguments[i]
+		avs[i] = esValue{id: an, value: arg}
+	}
+	e.s.cs = append(e.s.cs, make([]esValue, len(args)))
```
nickeskov, line 434: Change `len(args)` to len(avs)`. I think it will be more idiomatic than previous version.

pkg/ride/tree_evaluator.go
```diff
@@ -285,40 +399,86 @@ func (e *treeEvaluator) evaluateNativeFunction(name string, arguments []Node) (r
...
+	}
+	var tmp int
+	tmp, e.s.cl = e.s.cl, cl
+
+	r, err := e.walk(uf.Body)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to evaluate function '%s' body", name)
+	}
+	e.s.cs = e.s.cs[:len(e.s.cs)-1]
+	e.s.cl = tmp
```
nickeskov, line 446: Move it to defer to restore previous values in case of error.

pkg/ride/tree_evaluator.go
```diff
@@ -285,40 +399,86 @@ func (e *treeEvaluator) evaluateNativeFunction(name string, arguments []Node) (r
...
+	avs := make([]esValue, len(args))
+	for i, arg := range args {
+		an := uf.Arguments[i]
+		avs[i] = esValue{id: an, value: arg}
+	}
+	e.s.cs = append(e.s.cs, make([]esValue, len(args)))
+	for i, av := range avs {
+		e.s.cs[len(e.s.cs)-1][i] = av
+	}
+	var tmp int
```
nickeskov, line 438: I think `tmp` is not good naming.

pkg/ride/tree_evaluator.go
```diff
@@ -285,40 +399,86 @@ func (e *treeEvaluator) evaluateNativeFunction(name string, arguments []Node) (r
...
+	}
+	avs := make([]esValue, len(args))
+	for i, arg := range args {
+		an := uf.Arguments[i]
+		avs[i] = esValue{id: an, value: arg}
+	}
+	e.s.cs = append(e.s.cs, make([]esValue, len(args)))
+	for i, av := range avs {
+		e.s.cs[len(e.s.cs)-1][i] = av
+	}
```
nickeskov, line 437: Why you can't do it like that
```go
	e.s.cs = append(e.s.cs, avs)
```
instead of
```go
	e.s.cs = append(e.s.cs, make([]esValue, len(args)))
	for i, av := range avs {
		e.s.cs[len(e.s.cs)-1][i] = av
	}
```
?

pkg/ride/tree_evaluator.go
```diff
@@ -285,40 +399,86 @@ func (e *treeEvaluator) evaluateNativeFunction(name string, arguments []Node) (r
...
+	e.s.cl = tmp
+	return r, nil
+}
+
 func (e *treeEvaluator) walk(node Node) (rideType, error) {
 	switch n := node.(type) {
 	case *LongNode:
+		if e.cc.overflow() {
+			return nil, RuntimeError.New("evaluation complexity overflow")
+		}
```
nickeskov, line 455: Move this check before type switch

nickeskov (changes requested):

pkg/ride/functions_bigint.go
```diff
@@ -564,6 +564,37 @@ func medianListBigInt(_ environment, args ...rideType) (rideType, error) {
...
+	}
+	rp, ok := args[2].(rideInt)
+	if !ok {
+		return nil, errors.Errorf("sqrtBigInt: unexpected argument type '%s'", args[2].instanceOf())
+	}
+	round, err := roundingMode(args[3])
+	if err != nil {
+		return nil, errors.Wrap(err, "sqrtBigInt")
+	}
+	v := big.NewInt(0).Set(n.v)
```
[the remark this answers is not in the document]
alexeykiselev: Changed everywhere

pkg/ride/tree_evaluator.go
alexeykiselev: This context is unusable after error. No need to make deferred function bigger.

pkg/ride/tree_evaluator.go
alexeykiselev: Why not? It is a temporary place for the value that has to be restored later. What is your proposal for the name?

pkg/ride/tree_evaluator.go
alexeykiselev: Yep, my bad, it was iterative programming :)

pkg/ride/tree_evaluator.go
alexeykiselev: May be it's because there was no `avs` before.

nickeskov (approved):

PR #593, opened 2021-11-25
Changed the way of actions validation before ridev6

esuwu:


pkg/ride/environment.go
```diff
@@ -751,6 +753,28 @@ func (ws *WrappedState) incrementInvCount() {
...
+		)
+		if key.asset.Present {
+			balance, err = ws.diff.state.NewestAssetBalance(address, key.asset.ID)
+		} else {
+			balance, err = ws.diff.state.NewestWavesBalance(address)
+		}
+		if err != nil {
+			return err
+		}
+		if int64(balance)+balanceDiff.regular < 0 {
```
alexeykiselev, line 771: Here better to check for int64 overflow, as it made in state.

alexeykiselev (changes requested):

pkg/ride/environment.go
esuwu: check for an overflow and negative balance, right?

alexeykiselev (approved):

nickeskov (approved):

PR #632, opened 2021-12-22
Ride v6 disallow duplicate functions

alexeykiselev:


pkg/ride/tree_estimation_test.go
```diff
@@ -394,6 +368,7 @@ func TestNativeFoldEstimation(t *testing.T) {
 }
 
 func TestBrokenNativeFold(t *testing.T) {
+	// Hacked script with replaced fold function parameter
```
nickeskov, line 371: Duplicate

pkg/ride/tree_estimation_test.go
```diff
@@ -406,3 +381,32 @@ func TestBrokenNativeFold(t *testing.T) {
...
+
+func TestDeclarationDuplication(t *testing.T) {
+	for _, test := range []struct {
+		comment string
+		source  string
+		allowV4 bool
+	}{
+		{`Bad compaction result`, "AAIFAAAAAAAAAA8IAhIAGgkKAmExEgNhMTAAAAALAQAAAAJhMAAAAAAGAQAAAAJhMQAAAAAJAQAAAAJhMAAAAAABAAAAAmEyAAAAAAkBAAAAAmExAAAAAAEAAAACYTMAAAAACQEAAAACYTIAAAAAAQAAAAJhNAAAAAAJAQAAAAJhMwAAAAABAAAAAmE1AAAAAAkBAAAAAmE0AAAAAAEAAAACYTYAAAAACQEAAAACYTUAAAAAAQAAAAJhNwAAAAAJAQAAAAJhNgAAAAABAAAAAmE4AAAAAAkBAAAAAmE3AAAAAAEAAAACYTkAAAAACQEAAAACYTgAAAAAAQAAAAJhMQAAAAAJAQAAAAJhOQAAAAAAAAABAAAAAWkBAAAABGNhbGwAAAAABAAAAAFyCQEAAAACYTEAAAAAAwkAAAAAAAACBQAAAAFyBQAAAAFyBQAAAANuaWwJAAACAAAAAQIAAAAkU3RyaWN0IHZhbHVlIGlzIG5vdCBlcXVhbCB0byBpdHNlbGYuAAAAAChy+pw=", false},
+		{`V5: func f() = {func f() = true; f()}; f()`, "BQoBAAAAAWYAAAAACgEAAAABZgAAAAAGCQEAAAABZgAAAAAJAQAAAAFmAAAAABMDgQM=", true},
+		{`V5: func f() = {func a() = false; 1 + 2}; func g() = {func a() = false; 3 + 5}; f() > g()`, "BQoBAAAAAWYAAAAACgEAAAABYQAAAAAHCQAAZAAAAAIAAAAAAAAAAAEAAAAAAAAAAAIKAQAAAAFnAAAAAAoBAAAAAWEAAAAABwkAAGQAAAACAAAAAAAAAAADAAAAAAAAAAAFCQAAZgAAAAIJAQAAAAFmAAAAAAkBAAAAAWcAAAAAl60UNw==", true},
```
nickeskov, line 393: I think it would be good idea to add negative test case with correct script.

pkg/ride/tree_estimatorV4.go
```diff
@@ -42,6 +42,7 @@ func (e *treeEstimatorV4) estimate() (int, int, map[string]int, error) {
 	max := 0
 	m := make(map[string]int)
 	for i := 0; i < len(e.tree.Functions); i++ {
+		e.scope.resetFunctions()
```
nickeskov, line 45: Why we should reset functions? I assume that every `FunctionDeclarationNode` creates new scope. Is it correct?

pkg/ride/tree_estimatorV4.go
```diff
@@ -59,6 +60,7 @@ func (e *treeEstimatorV4) estimate() (int, int, map[string]int, error) {
 	}
 	vc := 0
 	if e.tree.HasVerifier() {
+		e.scope.resetFunctions()
```
nickeskov, line 63: Why we should reset functions? I assume that verifier creates new scope. Is it correct?

nickeskov (changes requested):

pkg/ride/tree_estimatorV4.go
alexeykiselev: Nope, during the estimation we use one scope for all callables, so it's dirtiness was not an issue then we don't check duplications. To check duplication correctly we have to clean the scope for every callable.

pkg/ride/tree_estimatorV4.go
alexeykiselev: Same for verfiier, it's just a last callable if present.

pkg/ride/tree_estimation_test.go
alexeykiselev: It's already there, the last two is correct scripts that should not rise error. May be `allowV4` is a bad name. Will change it.

nickeskov (approved):

PR #651, opened 2022-01-20
Eth JSON ABI http route

nickeskov:


pkg/api/app_eth.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/wavesplatform/gowaves/pkg/proto"
+	"github.com/wavesplatform/gowaves/pkg/proto/ethabi"
+	"github.com/wavesplatform/gowaves/pkg/ride"
+	"github.com/wavesplatform/gowaves/pkg/state"
+)
+
+func (a *App) EthereumDAppMethods(addr proto.WavesAddress) (ethabi.MethodsMap, error) {
+	scriptInfo, err := a.state.ScriptInfoByAccount(proto.NewRecipientFromAddress(addr))
+	if err != nil {
+		if state.IsNotFound(err) {
```
alexeykiselev, line 14: Why use this method? Is it no enough to use `errors.Is()`? This method `IsNotFound` doesn't use `errors.Is` inside, but checks only "upper" type of the error, doesn't look inside.

pkg/api/node_api.go
```diff
@@ -458,20 +457,52 @@ func (a *NodeApi) stateHash(w http.ResponseWriter, r *http.Request) error {
...
+			id,
+		),
+	)
+}
+
+func (a *NodeApi) EthereumDAppABI(w http.ResponseWriter, r *http.Request) error {
+	s := chi.URLParam(r, "address")
+	addr, err := proto.NewAddressFromString(s)
+	if err != nil {
+		if invalidRune := findFirstInvalidRuneInBase58String(s); invalidRune != nil {
```
alexeykiselev, line 474: Why for report the invalid Base58 rune? Is it not enough to report that the address is invalid? I think that the `findFirstInvalidRuneInBase58String` is highly non-optimal, why not just lookup for chars from Base58 alphabet?

alexeykiselev (changes requested):

pkg/api/node_api.go
nickeskov: Because we have to support scala's errors. Ok, I'll optimize it.

alexeykiselev (approved):

pkg/api/app_eth.go
nickeskov: State errors were refactored in last two commits.

PR #656, opened 2022-01-24
Eth invoke expression tx

esuwu:


pkg/state/transaction_performer.go
```diff
@@ -359,10 +359,14 @@ func (tp *transactionPerformer) performEthereumTransactionWithProofs(transaction
...
 	}
-	if _, ok := ethTx.TxKind.(*proto.EthereumInvokeScriptTxKind); ok {
-		if err := tp.stor.commitUncertain(info.blockID); err != nil {
-			return errors.Wrap(err, "failed to commit invoke changes")
-		}
+	_, isEthInvokeScript := ethTx.TxKind.(*proto.EthereumInvokeScriptTxKind)
+	_, isEthInvokeExpression := ethTx.TxKind.(*proto.EthereumInvokeExpressionTxKind)
+	if !(isEthInvokeScript || isEthInvokeExpression) {
+		return errors.New("performing transaction was called, but the transaction is not InvokeScript or InvokeExpression")
+	}
```
nickeskov, line 366: Use switch case

nickeskov (changes requested):

pkg/state/transaction_performer.go
nickeskov, line 366: These changes break compatibility with old logic

nickeskov (changes requested):

PR #657, opened 2022-01-25
Eth invoke and ethereum tx refactoring

esuwu:


pkg/state/script_caller.go
```diff
@@ -226,75 +226,24 @@ func (a *scriptCaller) invokeFunction(tree *ride.Tree, tx proto.Transaction, inf
...
-		env.ChooseSizeCheck(tree.LibVersion)
-		env.ChooseTakeString(info.rideV5Activated)
-		env.ChooseMaxDataEntriesSize(info.rideV5Activated)
-
-		// Since V5 we have to create environment with wrapped state to which we put attached payments
-		if tree.LibVersion >= 5 {
-			env, err = ride.NewEnvironmentWithWrappedState(env, payments, sender, info.rideV6Activated)
+	switch invokeType := invokeUnion.(type) {
+	case *proto.InvokeScriptTxUnion:
+		switch subTx := invokeType.SubTx.(type) {
```
nickeskov, line 231: I think it's good idea to move all code from this case to new function.

pkg/state/script_caller.go
```diff
@@ -307,76 +256,82 @@ func (a *scriptCaller) invokeFunction(tree *ride.Tree, tx proto.Transaction, inf
...
+		}
+		r, err = ride.CallFunction(env, tree, functionName, functionArguments)
+		if err != nil {
+			if appendErr := a.appendFunctionComplexity(ride.EvaluationErrorSpentComplexity(err), scriptAddress, functionName, defaultFunction, info); appendErr != nil {
+				return nil, appendErr
+			}
+			return nil, err
+		}
+	case *proto.InvokeExpressionTxUnion:
+		switch subTx := invokeType.SubTx.(type) {
```
nickeskov, line 294: I think it's good idea to move all code from this case to new function.

nickeskov (changes requested):

nickeskov (approved):
Almost all issues were fixed, approving.

PR #660, opened 2022-01-28
Internal payments validatin height updated. Tests on new validations …

alexeykiselev:
…added. Exclusions added.


pkg/state/invoke_applier.go
```diff
@@ -339,8 +339,8 @@ func (ia *invokeApplier) fallibleValidation(tx proto.Transaction, info *addlInvo
 		KeySizeValidationVersion: keySizeValidationVersion,
 		MaxDataEntriesSize:       maxDataEntriesSize,
 	}
-
-	if err := proto.ValidateActions(info.actions, restrictions, int(info.libVersion)); err != nil {
+	validatePayments := info.checkerInfo.height > ia.settings.InternalInvokePaymentsValidationAfterHeight
```
esuwu, line 342: Don't you think that we have to validate payments before `the hight` and after it? It seems to me that this code makes validation work only after `ia.settings.InternalInvokePaymentsValidationAfterHeight`. Maybe we should do something like this:
```
validatePayments := !(info.checkerInfo.height == ia.settings.InternalInvokePaymentsValidationAfterHeight)
```
Or is it not related to the last problem?

esuwu: I realized that before `the height` the payments has not been validated by the main Scala node, so they must be validated only after `the height`

nickeskov (approved):

PR #676, opened 2022-03-01
Fix complexity limit

alexeykiselev:


pkg/ride/environment.go
```diff
@@ -1158,6 +1164,10 @@ func (ws *WrappedState) ApplyToState(actions []proto.ScriptAction, env environme
 	return actions, nil
 }
 
+func isPositiveJVMLong(x int64) bool {
+	return x >= 0 && uint64(x) <= proto.MaxLongValue
```
nickeskov, line 1168: `uint64(x) <= proto.MaxLongValue` check is redundant (this expression will be true only if `x < 0` as `int64`).

nickeskov (changes requested):

PR #687, opened 2022-03-17
Allow mine with scripted account after feature 17 (RideV6).

nickeskov:


pkg/state/scripts_storage.go
```diff
@@ -344,15 +344,15 @@ func (ss *scriptsStorage) newestAccountHasVerifier(addr proto.WavesAddress, filt
 		return script.HasVerifier(), nil
 	}
 	script, err := ss.newestScriptAstByKey(keyBytes, filter)
-	if err != nil {
+	if err != nil { // TODO: check error type
```
alexeykiselev, line 347: Don't remove this TODOs, actually there's possible two types of errors NotFound-like and state failures. First one must be swallowed and replaced by `false, nil`, the second have to be passed upward.

pkg/state/scripts_storage.go
```diff
@@ -344,15 +344,15 @@ func (ss *scriptsStorage) newestAccountHasVerifier(addr proto.WavesAddress, filt
...
 		return false, nil
 	}
 	return script.HasVerifier(), nil
 }
 
 func (ss *scriptsStorage) accountHasVerifier(addr proto.WavesAddress, filter bool) (bool, error) {
 	script, err := ss.scriptByAddr(addr, filter)
-	if err != nil {
+	if err != nil { // TODO: check error type
```
alexeykiselev, line 355: Get it back, please.

pkg/state/state.go
```diff
@@ -701,6 +697,12 @@ func (s *stateManager) NewestScriptPKByAddr(addr proto.WavesAddress) (crypto.Pub
 	return s.stor.scriptsStorage.newestScriptPKByAddr(addr, true)
 }
 
+func (s *stateManager) NewestAccountHasScript(addr proto.WavesAddress) (bool, error) {
+	// This function is used only from stateInfoProvider interface for consensus.Validator,
```
alexeykiselev, line 701: Convert comment to TODO

alexeykiselev (changes requested):

pkg/state/scripts_storage.go
alexeykiselev: my bad

pkg/state/scripts_storage.go
alexeykiselev: my bad

alexeykiselev (approved):

PR #688, opened 2022-03-19
Allowed empty data entry key for old invoke transactions.

nickeskov:


pkg/state/transaction_checker.go
```diff
@@ -1202,7 +1202,7 @@ func (tc *transactionChecker) checkDataWithProofs(transaction proto.Transaction,
 		return nil, err
 	}
 	utf16KeyLen := tx.Version == 1 && !isRideV6Activated
-	if err := tx.Entries.Valid(utf16KeyLen); err != nil {
+	if err := tx.Entries.Valid(true, utf16KeyLen); err != nil {
```
esuwu, line 1205: Shouldn't we also call here `IsProtobuf`? The logic would be the same, but there won't be any hard code

pkg/ride/runtime.go
```diff
@@ -367,4 +367,5 @@ type environment interface {
 	rideV6Activated() bool
 	internalPaymentsValidationHeight() uint64
 	maxDataEntriesSize() int
+	protobufTransaction() bool
```
esuwu, line 370: we'd better call this method `isProtobufTransaction`

esuwu (changes requested):

pkg/state/transaction_checker.go
nickeskov: Nope, for data transaction that's correct. Empty keys are forbidden for any data transaction version.

pkg/ride/runtime.go
nickeskov: Yep, I agree with you. But as you can see `EvaluationEnvironment` struct has `isProtobufTransaction` field. So, I can't call this method as you suggest.

esuwu: isProtobufTx

esuwu (approved):

pkg/ride/runtime.go
nickeskov: Fixed

esuwu (approved):

esuwu (approved):

esuwu (approved):

PR #690, opened 2022-03-20
Block info by height fix.

nickeskov:


pkg/proto/block.go
```diff
@@ -423,6 +422,25 @@ type Block struct {
 	Transactions Transactions `json:"transactions,omitempty"`
 }
 
+func (b *Block) MarshalJSON() ([]byte, error) {
```
alexeykiselev, line 425: Do we have test on block marshalling on API level? I think we have to check that there is no unnecessary levels in block's json.

alexeykiselev (changes requested):

pkg/proto/block.go
nickeskov: Yep, we have. See `blockFromJSONToJSON` function in `pkg/proto/block_test.go`.

alexeykiselev (approved):

pkg/proto/block.go
nickeskov: Fixed. Tests have been added.

alexeykiselev (approved):

PR #691, opened 2022-03-21
Forbid multisig for create alias transaction before RideV6 activation.

nickeskov:


pkg/proto/types.go
```diff
@@ -2087,6 +2087,10 @@ func (p *ProofsV1) Valid() error {
 	return nil
 }
 
+func (p *ProofsV1) MultiSigned() bool {
```
alexeykiselev, line 2090: What if there is an empty proofs list? Such transaction is also must be validated by verifier. 
I think better name for this function is `SimpleSignature` with the opposite meaning. And it have to check not just the length of proofs, but the position of signature (it must be 0) and the length of signature itself.

pkg/state/transaction_checker.go
```diff
@@ -1103,13 +1103,20 @@ func (tc *transactionChecker) checkCreateAliasWithProofs(transaction proto.Trans
...
 	}
-	if !activated {
+	if !smartAccountsIsActivated {
 		return nil, errors.New("SmartAccounts feature has not been activated yet")
 	}
+	rideV6IsActivated, err := tc.stor.features.newestIsActivated(int16(settings.RideV6))
+	if err != nil {
+		return nil, err
+	}
+	if tx.Proofs.MultiSigned() && !rideV6IsActivated {
```
alexeykiselev, line 1117: So we just decline not simple signed transactions before Ride V6 to match scala node?

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #697, opened 2022-03-24
Added transfer contructor

esuwu:


pkg/ride/functions_proto.go
```diff
@@ -1376,6 +1376,25 @@ func dataTransaction(_ *treeEvaluator, _ environment, args ...rideType) (rideTyp
 	return obj, nil
 }
 
+func transfer(_ *treeEvaluator, _ environment, args ...rideType) (rideType, error) {
+	if err := checkArgs(args, 9); err != nil {
```
alexeykiselev, line 1380: Copy-paste!!! Do you expect 9 arguments?

pkg/ride/generate/main.go
```diff
@@ -361,6 +363,7 @@ func evaluationCatalogueV3EvaluatorV1() map[string]int {
 	m["TransferSet"] = 0
 	m["ScriptTransfer"] = 0
 	m["ScriptResult"] = 0
+	// TODO add evaluation for Transfer constructor
```
alexeykiselev, line 366: Evaluation complexity of all constructors is 0, but you have to add it explicitly otherwise it will be equal to estimation complexity.

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #698, opened 2022-03-24
Added bytes property

esuwu:


esuwu:
ByteVector type does not need to have "bytes" property.

PR #718, opened 2022-04-22
Allow invocation of functions with rideUnit as argument type

alexeykiselev:


pkg/ride/converters.go
```diff
@@ -1110,11 +1110,13 @@ func updateAssetInfoWithProofsToObject(scheme byte, tx *proto.UpdateAssetInfoWit
 	return r, nil
 }
 
-func convertListArguments(args rideList) ([]rideType, error) {
+func convertListArguments(args rideList, check bool) ([]rideType, error) {
```
esuwu, line 1113: Don't you think that `check` should be named `rideV6Activated` as well here?

esuwu (approved):

PR #724, opened 2022-04-28
Added dynamic fee in SetScriptTransaction

Anton-Rampage:


pkg/state/fee_validation.go
```diff
@@ -36,6 +36,8 @@ var feeConstants = map[proto.TransactionType]uint64{
 	proto.InvokeExpressionTransaction: 5,
 }
 
+var SetScriptTransactionV6Fee uint64 = 1
```
alexeykiselev, line 39: May be change to `const`

pkg/state/fee_validation_test.go
```diff
@@ -271,3 +272,46 @@ func TestSponsorshipFeeReduction(t *testing.T) {
...
+
+func randomScript(size uint64) (proto.Script, error) {
+	var s proto.Script = make([]byte, size)
+	_, err := rand.Read(s[:])
+	if err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
```
alexeykiselev, line 284: Please, add test on zero bytes script, it's a valid use case, setting empty script removes script from account. According to scala implementations setting zero size script should
cost 1 fee unit.

alexeykiselev (approved):

pkg/state/fee_validation.go
```diff
@@ -140,6 +142,26 @@ func minFeeInUnits(params *feeValidationParams, tx proto.Transaction) (uint64, e
...
+		fee = SetScriptTransactionV6Fee
+		stx, ok := tx.(*proto.SetScriptWithProofs)
+		if !ok {
+			return 0, errors.New("failed to convert interface to SetScriptTransaction")
+		}
+
+		stxBytesForFee := len(stx.Script)
+		if stxBytesForFee < 0 {
+			panic(fmt.Sprintf("BUG, CREATE REPORT: setScriptTx bytes size (%d) must not be lower than zero", stxBytesForFee))
+		}
```
nickeskov, line 162: Actually, this check is redundant. Length can't physically be lower than zero.

nickeskov (changes requested):

nickeskov (approved):

PR #747, opened 2022-05-23
Add exclude list for fsm

Anton-Rampage:


esuwu (dismissed):

pkg/node/state_fsm/fsm_sync.go
```diff
@@ -35,6 +35,20 @@ func (c conf) Now(tm types.Time) conf {
 type noopWrapper struct {
 }
 
+var (
+	syncSkipMessageList = proto.PeerMessageIDs{
+		proto.ContentIDGetBlock,
```
alexeykiselev, line 40: GetBlock and GetBlockIds messages should be allowed to enable sync after Protobuf blocks activation.

pkg/node/state_fsm/fsm_idle.go
```diff
@@ -19,6 +19,25 @@ type InvRequester interface {
 	Request(p types.MessageSender, id []byte)
 }
 
+var (
+	idleSkipMessageList = proto.PeerMessageIDs{
+		proto.ContentIDGetSignatures,
```
alexeykiselev, line 24: GetSignatures, GetBlock, GetBlockIds messages should be allowed to enable blocks distribution from idling nodes.

alexeykiselev (changes requested):

nickeskov (approved):

PR #755, opened 2022-05-27
Check on in-memory mapping of account on it's number added. Check on …

alexeykiselev:
…iterator removed from function as unnecessary. Function renamed to newwesEnrtyExists.


pkg/state/accounts_data_storage.go
```diff
@@ -274,31 +274,16 @@ func (s *accountsDataStorage) retrieveEntries(addr proto.Address, filter bool) (
...
-		iter.Release()
-		if err := iter.Error(); err != nil {
-			zap.S().Fatalf("Iterator error: %v", err)
-		}
-	}()
-
-	for iter.Next() {
-		return true, nil
-	}
-	return false, nil
```
nickeskov, line 301: Did you remove this because we also check state in `newestAddrToNum`?

nickeskov (dismissed):

nickeskov (approved):

PR #758, opened 2022-05-30
Corrected Ethereum transaction type

alexeykiselev:


esuwu (dismissed):

pkg/proto/eth_transaction.go
```diff
@@ -564,3 +627,13 @@ func (tx *EthereumTransaction) Protected() bool {
 		return true
 	}
 }
+
+func (tx *EthereumTransaction) UnmarshalJSON(data []byte) error {
+	tmp := struct {
+		Data HexBytes `json:"bytes"`
```
nickeskov, line 633: Rename this field to `Bytes`. It looks slightly weird when naming is different.

nickeskov (changes requested):

pkg/proto/eth_transaction.go
nickeskov: Fixed

nickeskov (approved):

PR #766, opened 2022-06-03
Add docker env for integration tests

Anton-Rampage:


Dockerfile.gowaves-it
```diff
@@ -0,0 +1,37 @@
+FROM golang:1.18.2-alpine3.15 as parent
```
nickeskov, line 1: Use `golang:1.18.3-alpine3.16`

nickeskov (changes requested):

itests/api/client.go
```diff
@@ -0,0 +1,108 @@
+package api
```
alexeykiselev, line 1: Why separate client? May be add those functions to common client in `pkg/client`?

alexeykiselev (changes requested):

itests/docker/docker.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil, err
+	}
+
+	logfile, err := os.Create(filepath.Clean(filepath.Join(pwd, logDir, goNodeLogFileName)))
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		_ = d.pool.Client.Logs(dc.LogsOptions{
```
nickeskov, line 148: When this goroutine will be stopped? Does this method return error?

itests/net/connection.go
```diff
@@ -0,0 +1,53 @@
+package net
+
+import (
```
nickeskov, line 3: Sort imports, please

itests/net/connection.go
```diff
@@ -0,0 +1,53 @@
...
+	"bufio"
+	"fmt"
+	"github.com/wavesplatform/gowaves/pkg/proto"
+	"net"
+	"time"
+)
+
+type OutgoingPeer struct {
+	conn net.Conn
+}
```
nickeskov, line 13: I can't find `Close` method for this struct. Do you close connections after using?

pkg/client/debug.go
```diff
@@ -271,3 +272,24 @@ func (a *Debug) StateChanges(ctx context.Context, id crypto.Digest) (*StateChang
...
 	return &out, response, nil
 }
+
+func (a *Debug) PrintMsg(ctx context.Context, msg string) (*Response, error) {
+	url, err := joinUrl(a.options.BaseUrl, "/debug/print")
+	if err != nil {
+		return nil, err
+	}
+
+	bts, err := json.Marshal(map[string]interface{}{"message": msg})
```
nickeskov, line 282: Create local struct for that

nickeskov (changes requested):

nickeskov (approved):

alexeykiselev (approved):

PR #767, opened 2022-06-03
add update timestamps of peers connections in SpawnOutgoingConnections

artemreyt:


pkg/node/peer_manager/storage/cbor.go
```diff
@@ -114,10 +114,6 @@ func (bs *CBORStorage) AddKnown(known []KnownPeer) error {
 
 	// Save existing known peers with their last attempt timestamps in backup
 	backup := bs.unsafeKnownIntersection(known)
-	// Fast path if all known peers already in storage
-	if len(backup) == len(known) {
-		return nil
-	}
```
nickeskov, line 120: Please, add tests which check the correctness of this changing.

nickeskov (changes requested):

pkg/node/peer_manager/storage/cbor.go
artemreyt: done

nickeskov (approved):

alexeykiselev (approved):

PR #770, opened 2022-06-06
Revert "forbid zero payments in invokes (#754)"

artemreyt:
This reverts commit 3eb62146ea6ea7e70cf30c3c609ada60eb5fb8c1.

### DIFF
--- a/pkg/ride/tree_evaluation_test.go
+++ b/pkg/ride/tree_evaluation_test.go
@@ -10228,7 +10228,7 @@ func TestInvokeDappAttachedPaymentsLimitAfterV6(t *testing.T) {
 	require.NotNil(t, res)
 }
 
-func TestZeroPaymentsV6(t *testing.T) {
+func TestInvokeDappFromDappWithZeroPayments(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/ride/tree_evaluation_test.go
```diff
@@ -10227,178 +10227,3 @@ func TestInvokeDappAttachedPaymentsLimitAfterV6(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, res)
 }
-
-func TestZeroPaymentsV6(t *testing.T) {
```
alexeykiselev, line 10231: May be better not to remove the test, but remove the last check? If there is no test duplication.

alexeykiselev (changes requested):

pkg/ride/tree_evaluation_test.go
artemreyt: Yep, changed name to more appropriate and left checks for both cases rideV6Activated=false/true.

alexeykiselev (approved):

PR #787, opened 2022-06-22
Fixed check on using of invocation functions in verifier. Tests added.

alexeykiselev:


pkg/ride/tree_estimatorV3.go
```diff
@@ -9,6 +9,7 @@ import (
 type fd struct {
 	cost   int
 	usages []string
+	dirty  bool
```
esuwu, line 12: Could you please add a comment what `dirty` means in this context or change the name of the variable

esuwu (changes requested):

pkg/ride/tree_estimatorV4.go
```diff
@@ -96,130 +102,132 @@ func (e *treeEstimatorV4) wrapFunction(node *ast.FunctionDeclarationNode) ast.No
...
-			return res, nil
+			return res, ci || li, nil
 		}
 		res, err := common.AddInt(ce, re)
 		if err != nil {
-			return 0, err
+			return 0, false, err
 		}
-		return res, nil
+		return res, ci || ri, nil
```
esuwu, line 138: Would be also nice, if we add a comment above about `ci`, `ri`, `li`, `le`, `ce` explaining abbreviation

pkg/ride/tree_estimatorV4.go
```diff
@@ -96,130 +102,132 @@ func (e *treeEstimatorV4) wrapFunction(node *ast.FunctionDeclarationNode) ast.No
 	return block
 }
 
-func (e *treeEstimatorV4) walk(node ast.Node, enableInvocation bool) (int, error) {
+func (e *treeEstimatorV4) walk(node ast.Node) (int, bool, error) {
```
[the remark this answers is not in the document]
alexeykiselev: I think it is not required yet, too few results. I'll add comment to explain results.

pkg/ride/tree_estimatorV4.go
alexeykiselev: Ok, I'll try, but they so local, so the short name is quite ok.

Issue #270, opened 2020-03-17
[FEATURE] Use TestNet scheme in tests


alexeykiselev:
**Аbstract**
Now in tests used a mixture of MainNet and TestNet schemes. Better to stick to TestNet. Some tests that checks real objects from MainNet should use scheme accordingly. 

**Motivation and Purposes**
Simplifying the decision which scheme to use while testing.

cryptopragmatic:
@alexeykiselev, I've changed many of MainNetScheme to TestNetScheme, most of the ones that just needed changing without any additional refactoring in a way that it doesn't break tests.

```
find . -type f -name "*_test.go"  | xargs grep "MainNet" |  wc -l
943
git checkout tests-testnet 
find . -type f -name "*_test.go"  | xargs grep "MainNet" |  wc -l
261
```
```
find . -type f -name "*_test.go"  | xargs grep "MainNetScheme" |  wc -l
145
git checkout master
find . -type f -name "*_test.go"  | xargs grep "MainNetScheme" |  wc -l
821
```

Pull request is here:

https://github.com/wavesplatform/gowaves/pull/790

PR #790, opened 2022-06-26
Mainnet -> testnet in tests

cryptopragmatic:
As suggested in issue #270, I've changed many of MainNetScheme to TestNetScheme, most of the ones that just needed changing without any additional refactoring in a way that it doesn't break tests.

```
find . -type f -name "*_test.go"  | xargs grep "MainNet" |  wc -l
943
git checkout tests-testnet 
find . -type f -name "*_test.go"  | xargs grep "MainNet" |  wc -l
261
```
```
find . -type f -name "*_test.go"  | xargs grep "MainNetScheme" |  wc -l
145
git checkout master
find . -type f -name "*_test.go"  | xargs grep "MainNetScheme" |  wc -l
821
```


nickeskov:
@cryptopragmatic Thanks for your contribution!

PR #800, opened 2022-07-07
Correct complexity calculator selection for scripts of Ride versions …

alexeykiselev:
…prior 5 added

### DIFF
--- a/pkg/ride/tree_evaluation_test.go
+++ b/pkg/ride/tree_evaluation_test.go
@@ -2569,7 +2569,7 @@ func TestInvokeDAppFromDAppScript6(t *testing.T) {
 	wrappedSt = *NewWrappedSt
 
 	_, err = CallFunction(env, tree, "foo", proto.Arguments{})
-	require.Error(t, err)
+	assert.EqualError(t, err, "invoke: too many internal invocations")
 
 	expectedDiffResult := initWrappedState(smartState(), env, tree.LibVersion).diff
 	assert.Equal(t, expectedDiffResult.data, wrappedSt.diff.data)
@@ -2741,7 +2741,7 @@ func TestReentrantInvokeDAppFromDAppScript6(t *testing.T) {
 	wrappedSt = *NewWrappedSt
 
 	_, err = CallFunction(env, tree, "foo", proto.Arguments{})
-	require.Error(t, err)
+	assert.EqualError(t, err, "reentrantInvoke: too many internal invocations")
 
 	expectedDiffResult := initWrappedState(smartState(), env, tree.LibVersion).diff
 	assert.Equal(t, expectedDiffResult.data, wrappedSt.diff.data)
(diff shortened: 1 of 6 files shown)


pkg/ride/tree_evaluation_test.go
```diff
@@ -2568,29 +2568,10 @@ func TestInvokeDAppFromDAppScript6(t *testing.T) {
...
-		Transfers:    make([]*proto.TransferScriptAction, 0),
-		Issues:       make([]*proto.IssueScriptAction, 0),
-		Reissues:     make([]*proto.ReissueScriptAction, 0),
-		Burns:        make([]*proto.BurnScriptAction, 0),
-		Sponsorships: make([]*proto.SponsorshipScriptAction, 0),
-		Leases:       make([]*proto.LeaseScriptAction, 0),
-		LeaseCancels: make([]*proto.LeaseCancelScriptAction, 0),
-	}
-	assert.Equal(t, expectedActionsResult, sr)
+	_, err = CallFunction(env, tree, "foo", proto.Arguments{})
```
esuwu, line 2571: Why the function with a recursive call should return an error? I see you changed the number of invocation count, but how is that related with a recursive call?

alexeykiselev: Because it should return error in case of exceeding recursion limit

esuwu (approved):

pkg/ride/tree_evaluation_test.go
```diff
@@ -2568,29 +2568,10 @@ func TestInvokeDAppFromDAppScript6(t *testing.T) {
...
+	require.Error(t, err)
```
nickeskov, line 2572: Please, add assertion with error message to make sure that script has been failed with `invoke: too many internal invocations` reason

pkg/ride/tree_evaluation_test.go
```diff
@@ -2760,29 +2740,10 @@ func TestReentrantInvokeDAppFromDAppScript6(t *testing.T) {
...
```
nickeskov, line 2744: The same issue as above

pkg/state/script_caller.go
```diff
@@ -207,7 +207,7 @@ func (a *scriptCaller) callAssetScript(tx proto.Transaction, assetID crypto.Dige
 }
 
 func (a *scriptCaller) invokeFunction(tree *ast.Tree, tx proto.Transaction, info *fallibleValidationParams, scriptAddress proto.WavesAddress) (ride.Result, error) {
-	env, err := ride.NewEnvironment(a.settings.AddressSchemeCharacter, a.state, a.settings.InternalInvokePaymentsValidationAfterHeight)
+	env, err := ride.NewEnvironment(a.settings.AddressSchemeCharacter, a.state, a.settings.InternalInvokePaymentsValidationAfterHeight, info.blockV5Activated, info.rideV5Activated)
```
nickeskov, line 210: The last argument should be ` info.rideV6Activated`, not ` info.rideV5Activated`.

nickeskov (changes requested):

nickeskov (approved):

PR #813, opened 2022-07-14
Initialize all fields in 'ActionsValidationRestrictions' struct.

nickeskov:
Add `ActionsValidationRestrictions.Scheme` and `ActionsValidationRestrictions.ScriptAddress` fields initialisation.
Moved `env.setNewDAppAddress` call after caller's attached payments applying.


pkg/proto/actions_validator.go
```diff
@@ -124,13 +123,9 @@ func ValidateAttachedPaymentScriptAction(action *AttachedPaymentScriptAction, re
 		return errors.New("negative transfer amount")
 	}
 	if restrictions.DisableSelfTransfers {
-		senderAddress := restrictions.ScriptAddress
-		if action.SenderPK() != nil {
```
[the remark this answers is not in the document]
nickeskov: Fixed

artemreyt (dismissed):

pkg/ride/functions_proto.go
```diff
@@ -194,6 +189,12 @@ func performInvoke(invocation invocation, env environment, args ...rideType) (ri
 		return nil, err
 	}
 
+	address, err := env.state().NewestRecipientToAddress(recipient)
```
esuwu, line 192: Why did you decide to move setting new address?

pkg/ride/functions_proto.go
```diff
@@ -178,11 +178,6 @@ func performInvoke(invocation invocation, env environment, args ...rideType) (ri
 			Asset:     payment.Asset,
 		}
 	}
-	address, err := env.state().NewestRecipientToAddress(recipient)
```
esuwu, line 181: We need to have script addess be set before applying payments, before the function `getLibVersion` expects to see script address in ws.callee()
So, it's better not to move setting address from here

esuwu (changes requested):

esuwu (dismissed):

PR #821, opened 2022-07-20
Improve blockcmp

alexeykiselev:


cmd/blockcmp/internal/entities.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (a *Issue) Equal(b Issue) bool {
+	return a.AssetID == b.AssetID && a.Name == b.Name && a.Description == b.Description &&
+		a.Amount == b.Amount && a.Decimals == b.Decimals && a.Reissuable == b.Reissuable &&
+		a.Script == b.Script && a.Nonce == b.Nonce
+}
+
+func (a *Issue) String() string {
+	return fmt.Sprintf(
```
nickeskov, line 79: I think `reflect` can be helpful here.

cmd/blockcmp/internal/entities.go
```diff
@@ -0,0 +1,243 @@
...
+	Description string
+	Amount      int
+	Decimals    int
+	Reissuable  bool
+	Script      string
+	Nonce       int
+}
+
+func (a *Issue) Equal(b Issue) bool {
+	return a.AssetID == b.AssetID && a.Name == b.Name && a.Description == b.Description &&
```
nickeskov, line 73: I think `reflect` can be helpful here.

nickeskov (changes requested):

cmd/blockcmp/internal/entities.go
alexeykiselev: May be later

cmd/blockcmp/internal/entities.go
alexeykiselev: May be later :)

nickeskov (approved):

PR #833, opened 2022-07-27
Proto version refactoring.

nickeskov:


cmd/retransmitter/retransmit/network/incoming_peer.go
```diff
@@ -66,7 +66,7 @@ func RunIncomingPeer(ctx context.Context, params IncomingPeerParams) {
 	writeHandshake := proto.Handshake{
 		AppName: params.WavesNetwork,
 		// pass the same minor version as received
-		Version:      proto.Version{Major: 0, Minor: readHandshake.Version.Minor, Patch: 0},
```
alexeykiselev, line 69: Looks very strange. Why we always pass 0 as major version? I think we should only drop patch part from the incoming version.

cmd/binclient/main.go
```diff
@@ -55,7 +52,7 @@ func main() {
 		zap.S().Fatal("please, provide 'version' CLI argument")
 	}
 
-	version, err := parseVersion(*version)
```
alexeykiselev, line 58: Those variables overwrites looks terrible

alexeykiselev (changes requested):

cmd/retransmitter/retransmit/network/incoming_peer.go
nickeskov: I agree. Fixed

cmd/binclient/main.go
nickeskov: Also fixed

alexeykiselev (approved):

PR #839, opened 2022-08-01
add struct generaration for rideObject

Anton-Rampage:


alexeykiselev (dismissed):

pkg/ride/generate/internal/coder.go
```diff
@@ -0,0 +1,58 @@
...
+		for _, imp := range c.imports {
+			sb.WriteString(imp + "\n")
+		}
+		sb.WriteString(")\n")
+	}
+	sb.WriteString(c.sb.String())
+	code := sb.String()
+	b, err := format.Source([]byte(code))
+	if err != nil {
+		println("====== Generated code ======")
```
nickeskov, line 52: Use `fmt.Println` instead of `println`.

alexeykiselev (approved):

PR #851, opened 2022-08-09
add matches and keys to AddressesData

bodrych:


pkg/client/addresses.go
```diff
@@ -320,14 +321,46 @@ func (a *Addresses) BalanceAfterConfirmations(
...
+
+type AddressesDataParam func(*AddressesDataParams)
+
+func WithMatches(matches string) AddressesDataParam {
+	return func(p *AddressesDataParams) {
+		p.matches = matches
+	}
+}
+
+func WithKeys(keys []string) AddressesDataParam {
```
alexeykiselev, line 337: May be better replace `[]string` with `...string` and later do like this `p.keys = []string(keys)`. The usage will be easier.

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #872, opened 2022-08-25
add parallel testing

Anton-Rampage:


alexeykiselev (dismissed):

irina-pereiaslavskaia (approved):

Makefile
```diff
@@ -39,7 +39,7 @@ gotest-coverage:
 itest:
 	mkdir -p build/config
 	mkdir -p build/logs
-	go test $$(go list ./... | grep "/itests")
+	go test -parallel 3 $$(go list ./... | grep "/itests")
```
nickeskov, line 42: I think `parallel` in CI should be `1`. Create new `itest-ci` target for CI.

itests/docker/docker.go
```diff
@@ -100,18 +109,31 @@ func removeExistsContainers(pool *dockertest.Pool) error {
...
 
-func (d *Docker) RunContainers(ctx context.Context, paths config.ConfigPaths) error {
-	goNodeRes, err := d.runGoNode(ctx, paths.GoConfigPath)
+func (d *Docker) RunContainers(ctx context.Context, paths config.ConfigPaths, suiteName string) (*Ports, error) {
+	pwd, err := os.Getwd()
 	if err != nil {
-		return err
+		return nil, err
+	}
+	err = os.MkdirAll(filepath.Clean(filepath.Join(pwd, logDir, suiteName)), os.ModePerm)
```
nickeskov, line 117: `filepath.Join` returns cleaned path. `filepath.Clean` here is unnecessary.

itests/integration_test.go
```diff
@@ -32,19 +33,19 @@ func (suite *ItestSuite) SetupSuite() {
 	suite.NoError(err, "couldn't create config")
 	suite.cfg = cfg
 
-	docker, err := d.NewDocker()
+	suiteName := strings.ToLower(suite.T().Name())
```
nickeskov, line 36: Why `ToLower` is necessary here?

itests/integration_test.go
```diff
@@ -82,6 +83,17 @@ func (suite *ItestSuite) Test_SendTransaction() {
...
+
+func TestItest2Suite(t *testing.T) {
+	t.Parallel()
+	suite.Run(t, new(ItestSuite))
+}
+
+func TestItest3Suite(t *testing.T) {
+	t.Parallel()
 	suite.Run(t, new(ItestSuite))
 }
```
nickeskov, line 99: I think these tests can be replaced by one test with subtests. Take a look:
```go
func TestItestParallel(t *testing.T) {
	for _, name := range []string{"1", "2", "3"} {
		t.Run(name, func(t *testing.T) {
			t.Parallel()
			suite.Run(t, new(ItestSuite))
		})
	}
}
```

itests/node_client/node_client.go
```diff
@@ -28,6 +29,13 @@ type NodesClients struct {
 	ScalaClients *NodeClients
 }
 
+func NewNodesClients(t *testing.T, ports *d.Ports) *NodesClients {
+	return &NodesClients{
+		GoClients:    NewNodeClient(t, ports.Scala.RestApiPort, ports.Scala.GrpcPort),
+		ScalaClients: NewNodeClient(t, ports.Go.RestApiPort, ports.Go.GrpcPort),
```
nickeskov, line 35: Use scala ports config for `ScalaClients` (the same for `GoClients`).

nickeskov (changes requested):

itests/integration_test.go
alexeykiselev: I think `ToLower` makes every string better

itests/integration_test.go
alexeykiselev: Human readability is not good here

itests/integration_test.go
nickeskov: I think the reason for `ToLower` here is different :)

Anton-Rampage: It necessary because docker doesn't work with uppercase in the container name

itests/integration_test.go
Anton-Rampage: Slash appears in the name of the container and the docker doesn't work again

nickeskov (approved):

PR #873, opened 2022-08-30
Checking that the correct blockchain is being loaded

Anton-Rampage:


pkg/state/state.go
```diff
@@ -486,6 +487,23 @@ func newStateManager(dataDir string, amend bool, params StateParams, settings *s
 		if err := state.applyPreActivatedFeatures(settings.PreactivatedFeatures, settings.Genesis.BlockID()); err != nil {
 			return nil, errors.Wrap(err, "failed to apply pre-activated features")
 		}
+	} else {
```
nickeskov, line 490: Unnecessary `else`.

pkg/state/state.go
```diff
@@ -486,6 +487,23 @@ func newStateManager(dataDir string, amend bool, params StateParams, settings *s
...
+		}
+		genesisBts, err := genesis.MarshalBinary()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to marshal genesis from state")
+		}
+		genesisFromConfigBts, err := settings.Genesis.MarshalBinary()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to marshal genesis from config")
+		}
+		if !bytes.Equal(genesisBts, genesisFromConfigBts) {
```
nickeskov, line 504: You can compare `BlockID`s instead of binary representation.

pkg/state/state.go
```diff
@@ -486,6 +487,23 @@ func newStateManager(dataDir string, amend bool, params StateParams, settings *s
...
+			return nil, fmt.Errorf("genesis blocks from state and config mismatch")
```
nickeskov, line 505: We use `errors.Errorf`

nickeskov (changes requested):

nickeskov (approved):

PR #881, opened 2022-09-08
Fix transactions validation

nickeskov:


pkg/state/state.go
```diff
@@ -585,8 +583,7 @@ func (s *stateManager) addGenesisBlock() error {
 	if _, err := s.stor.putStateHash(nil, 1, s.genesis.BlockID()); err != nil {
 		return err
 	}
-	verifyError := <-chans.errChan
-	if verifyError != nil {
+	if verifyError := chans.closeAndWait(); verifyError != nil {
```
alexeykiselev, line 586: I think there is an incorrect order of operations. Why for apply and store diff and state hashes if `verifyError` happened? May be move `closeAndWait` up to where `close` was?

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #902, opened 2022-10-03
Add 'enable-metamask-log' command line argument for node.

nickeskov:


cmd/node/node.go
```diff
@@ -479,6 +480,7 @@ func apiRunOptsFromCLIFlags() *api.RunOptions {
 	if *enableMetaMaskAPI {
 		if *buildExtendedApi {
 			opts.EnableMetaMaskAPI = *enableMetaMaskAPI
+			opts.EnableMetaMaskAPILog = *enableMetaMaskAPILog
```
esuwu, line 483: where is it used exactly? what is it for?

nickeskov: This option enables or disables metamask API request logging. Take a look
https://github.com/wavesplatform/gowaves/blob/0633158232b01d19b06376ab6e81de15f2efe9e6/pkg/api/routes.go#L139-L150

esuwu (approved):

PR #958, opened 2022-12-05
Log transaction broadcast.

nickeskov:


pkg/node/state_fsm/fsm_common.go
```diff
@@ -22,19 +23,42 @@ func newPeer(fsm FSM, p peer.Peer, peers peer_manager.PeerManager) (FSM, Async,
...
+func tryBroadcastTransaction(fsm FSM, baseInfo BaseInfo, p peer.Peer, t proto.Transaction) (_ FSM, _ Async, err error) {
+	defer func() {
+		if err != nil {
+			err = fsm.Errorf(proto.NewInfoMsg(err))
+		}
+	}()
+	if zap.S().Level() <= zap.DebugLevel {
+		defer func() {
+			if genIDErr := t.GenerateID(baseInfo.scheme); genIDErr != nil {
+				zap.S().Errorf("[%s] Failed to generate ID for transaction: %v", fsm.String(), genIDErr)
```
artemreyt, line 35: Is it okay that we get new ERROR messages if we just enable DEBUG level? Seems more error messages shouldn't appear if we just turn on debug-level.

pkg/node/state_fsm/fsm_common.go
```diff
@@ -22,19 +23,42 @@ func newPeer(fsm FSM, p peer.Peer, peers peer_manager.PeerManager) (FSM, Async,
...
+				return
+			}
+			txIDBytes, getIDErr := t.GetID(baseInfo.scheme)
+			if getIDErr != nil {
+				zap.S().Errorf("[%s] Failed to get ID for transaction: %v", fsm.String(), getIDErr)
```
artemreyt, line 40: same

pkg/node/state_fsm/fsm_common.go
```diff
@@ -22,19 +23,42 @@ func newPeer(fsm FSM, p peer.Peer, peers peer_manager.PeerManager) (FSM, Async,
...
+				return
+			}
+			txID := base58.Encode(txIDBytes)
+			if err != nil {
+				err = errors.Wrapf(err, "Failed to broadcast transaction %q", txID)
```
artemreyt, line 45: Here's similar, just here we change info message in case of debug mode. I think it's not good that we have different info messages in INFO and DEBUG mode

pkg/node/state_fsm/fsm_common.go
artemreyt: discussed, decided to leave as is.

pkg/node/state_fsm/fsm_common.go
artemreyt: discussed, decided to leave as is.

pkg/node/state_fsm/fsm_common.go
artemreyt: discussed, decided to leave as is.

artemreyt (approved):

PR #968, opened 2022-12-12
Fix client structs

artemreyt:


pkg/client/state_changes.go
```diff
@@ -21,8 +21,10 @@ type IssueAction struct {
 	Name           string        `json:"name"`
 	Description    string        `json:"description"`
 	Decimals       int32         `json:"decimals"`
+	Quantity       int64         `json:"quantity"`
 	Reissuable     bool          `json:"isReissuable"`
-	CompiledScript []byte        `json:"compiledScript"`
+	CompiledScript *[]byte       `json:"compiledScript"` // optional
```
nickeskov, line 26: Pointer here is unnecessary because zero value of any slice is `nil`.

nickeskov (changes requested):

pkg/client/state_changes_test.go
```diff
@@ -184,624 +217,65 @@ func TestUnmarshalStateChangesAllFields(t *testing.T) {
 	require.Equal(t, expectedStateChanges, actualChanges, "non equal stateChanges")
 }
 
-func TestName(t *testing.T) {
+func TestUnmarshalStateChangesLeaseCancelParams(t *testing.T) {
 	// mainnet tx 'HPsKrSJFFe9duhGbT3jb1YYnTninzmb99JRdRTtRefLw'
```
nickeskov, line 221: Is this comment still relevant?

pkg/client/state_changes.go
nickeskov: Oh, I've realized, that yes. Closing.

pkg/client/state_changes_test.go
nickeskov: Oh, I've realized, that yes. Closing.

pkg/client/state_changes.go
artemreyt: done

nickeskov (approved):

alexeykiselev (approved):

PR #972, opened 2022-12-14
Run itest on self hosted runner

alexeykiselev:


alexeykiselev:
Too many unnecessary changes have been made.

PR #978, opened 2022-12-23
Set microblock interval

alexeykiselev:


pkg/node/state_fsm/fsm_ng.go
```diff
@@ -161,7 +159,8 @@ func (a *NGFsm) MinedBlock(block *proto.Block, limits proto.MiningLimits, keyPai
 	a.baseInfo.actions.SendScore(a.baseInfo.storage)
 	a.baseInfo.CleanUtx()
 
-	return NewNGFsm12(a.baseInfo), Tasks(NewMineMicroTask(1*time.Second, block, limits, keyPair, vrf)), nil
+	// Try to mine micro-block just after key-block generation
+	return NewNGFsm12(a.baseInfo), Tasks(NewMineMicroTask(0, block, limits, keyPair, vrf)), nil
```
nickeskov, line 163: Please, can you remind me the main point why zero delay is fine here? We definitely have guarantees that this task will be started after `MinedBlock` ends, though all tasks come through `spawnAsync`function. So their launch time is undefined and can only be chosen by go runtime.

nickeskov (changes requested):

pkg/node/state_fsm/fsm_ng.go
alexeykiselev: This way we reproduce the scala-node behaviour. It generates first micro-block just after generation of key-block. The second and following micro-blocks generated with given interval.

nickeskov (approved):

PR #980, opened 2022-12-27
Improve wallet config

esuwu:


pkg/proto/common.go
```diff
@@ -159,6 +159,19 @@ func NetworkStrFromScheme(scheme Scheme) string {
 	return prefix + string(scheme)
 }
 
+func ParseSchemeFromStr(scheme string) (Scheme, error) {
```
alexeykiselev, line 162: This function is quite limited. We should generate wallets for integration tests with `L` scheme and so on. Better to receive one letter as cli option.

alexeykiselev (changes requested):

pkg/proto/common.go
nickeskov: Totally agree.

cmd/wallet/main.go
```diff
@@ -1,163 +1,378 @@
...
+	if exists(walletPath) {
+		fmt.Print("Wallet already exists on the provided path. Rewrite? Y/N: ")
+		var answer string
+		_, err := fmt.Scanf("%s", &answer)
 		if err != nil {
-			fmt.Printf("Err: %s\n", err.Error())
-			return
+			return errors.Wrap(err, "failed to get the answer on rewriting the existing wallet")
+		}
+		switch answer {
```
nickeskov, line 340: I think using case insensitive matching is a good idea.

cmd/wallet/main.go
```diff
@@ -1,163 +1,378 @@
...
+		case "Y":
+			wlt = wallet.NewWallet()
+		case "N":
+			return nil
+		default:
+			return errors.New("unknown command")
```
nickeskov, line 346: What if I'll press `enter` key multiple times? Is it an error? I don't think so. I think we should consider it like `N` (this option should be set by default).

cmd/wallet/main.go
```diff
@@ -1,163 +1,378 @@
...
+	}
+	fmt.Println(`Available options: 
+		 1: Generate new seed phrase and wallet
+		 2: Create a wallet based on an existing seed phrase. Requires seed phrase argument to be provided
+		 3: Create a wallet based on an existing base58-encoded seed phrase. Requires seed phrase argument to be provided and seed-base58 flag marked "true"
+		 4: Create a wallet based on an existing base58-encoded account seed. Requires account seed argument in Base58 format`)
+
+	var walletCredentials *WalletCredentials
+	var choice WalletGenerationChoice
+	_, err = fmt.Scanf("%s", &choice)
```
nickeskov, line 309: I don't see any message for user that option should be entered as a number.

nickeskov (dismissed):

cmd/wallet/main.go
```diff
@@ -1,163 +1,378 @@
...
 	"fmt"
 	"io/fs"
+	"log"
 	"os"
 	"os/user"
 	"path/filepath"
 
-	"github.com/howeyc/gopass"
 	"github.com/mr-tron/base58"
 	flag "github.com/spf13/pflag"
```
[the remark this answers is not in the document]
esuwu: Yes, but then it wouldn't be possible to do this `./wallet [command] <flags>. This library allows to use commands along with flags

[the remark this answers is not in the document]
alexeykiselev: I think we don't need commands for such a simple utiltiy

pkg/proto/common.go
esuwu: Even if receiving one letter, it must be `string` and we'll need to parse it into `proto.Scheme`. Otherwise, a user would have to enter `Uint` number like 87 ('W') and so on..

esuwu: Or we can implement flag.Value interface for `proto.Scheme` to make a custom flag for our specific case

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
 	opts := Opts{}
-
-	flag.BoolVarP(&opts.Force, "force", "f", false, "Overwrite existing wallet")
-	flag.StringVarP(&opts.PathToWallet, "wallet", "w", "", "Path to wallet")
-	flag.BoolVarP(&opts.Base58, "base58", "b", false, "Input seed as Base58 encoded string")
+	var scheme SchemeFlag
+	flag.BoolVar(&opts.New, newOpt, false, "Generate and add a new seed phrase (Primary flag)")
+	flag.BoolVar(&opts.New, showOpt, false, "Show existing wallet credentials (Primary flag)")
+	flag.StringVar(&opts.SeedPhrase, seedPhraseOpt, "", "Import a seed phrase (Primary flag)")
+	flag.BoolVar(&opts.IsSeedPhraseBase58, seedPhraseBase58Opt, false, "Import a base58-encoded seed phrase (Primary flag)")
```
alexeykiselev, line 89: Again, this flag that shows that given seed phrase is Base58 encoded is not primary. Because from one point of view, "primary flag" means that only one can be used at a time. Which is not the case for the `IsSeedPhraseBase58`. I think only `-new`,  `-seed-phrase` and `-account-seed` are mutually exclusive. If someone use `-seed-phrase-base58` with `-account-seed` it can be ignored, because account seed is always Base58 encoded.

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
-  show         Print wallet data
+Flags:
+`
 
+var examples = `
+Examples:
+	./wallet -show -wallet <wallet path> // Show existing wallet credentials
+	./wallet -new // Generate a seed phrase and a wallet 
+	./wallet -seed-phrase "..." // Import a seed phrase
+	./wallet -seed-phrase-base58 "..." // Import a Base58 encoded seed phrase
```
alexeykiselev, line 52: From example `-seed-phrase-base58` looks like a string option, but actually it is boolean.

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
+		return errors.Wrap(err, "failed to get the password to encode the account seed")
+	}
+
+	if len(pass) == 0 {
+		return errors.Wrap(err, "the password's length is zero")
+	}
+
+	var wlt wallet.Wallet
+	if exists(walletPath) {
+		fmt.Print("Wallet already exists on the provided path. Rewrite? Y/N: ")
```
alexeykiselev, line 346: This routine doesn't allow to add new seed to existing wallet. It only allows to overwrite it. The question should be "Wallet already exist. Do you want to add seed to existing wallet or overwrite it (existing seeds will be lost)? [A]dd/[O]verwrite/[C]ancel".

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
 
 	switch command {
-	case "add":
-		addToWallet(opts)
-	case "show":
-		show(opts)
+	case showOpt:
+		err = show(opts)
+		if err != nil {
+			log.Printf("failed to show wallet's credentials: %v", err)
```
alexeykiselev, line 119: Please, capitalise first letter of log messages.

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
+	flag.StringVar(&opts.Base58AccountSeed, accountSeedBase58Opt, "", "Import a base58-encoded account seed (Primary flag)")
+	flag.StringVar(&opts.PathToWallet, "wallet", "", "Path to the wallet file")
+	flag.IntVar(&opts.AccountNumber, "number", 0, "Account number. 0 is default")
+	flag.Var(&scheme, schemeOpt, "Network scheme: MainNet=W, TestNet=T, StageNet=S, CustomNet=E. MainNet is default")
 
 	flag.Parse()
 
-	command := flag.Arg(0)
+	possibleFlags := []string{newOpt, showOpt, seedPhraseOpt, seedPhraseBase58Opt, accountSeedBase58Opt}
+	if moreThanOnePrimaryFlagPassed(possibleFlags) {
```
alexeykiselev, line 98: All this flags to commands logic is overcomplicated. I think `moreThanOnePrimaryFlagPassed` check is enough.

alexeykiselev (changes requested):

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
 
 func main() {
 	opts := Opts{}
-
-	flag.BoolVarP(&opts.Force, "force", "f", false, "Overwrite existing wallet")
-	flag.StringVarP(&opts.PathToWallet, "wallet", "w", "", "Path to wallet")
-	flag.BoolVarP(&opts.Base58, "base58", "b", false, "Input seed as Base58 encoded string")
+	var scheme SchemeFlag
+	flag.BoolVar(&opts.New, newOpt, false, "Generate and add a new seed phrase (Primary flag)")
+	flag.BoolVar(&opts.New, showOpt, false, "Show existing wallet credentials (Primary flag)")
```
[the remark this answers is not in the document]
esuwu: "What stops me from creating new wallet and showing its content at once?" - the content is going to be shown anyway

cmd/wallet/main.go
esuwu: My fault, it's not supposed to be there, I thought I pushed the commit

cmd/wallet/main.go
esuwu: Once again, it's supposed to be string

cmd/wallet/main.go
esuwu: I wanted to have only one point of creating the possible flags, so if we ever add / remove some of them, we wouldn't have to change it in every function

esuwu: Moved it to global var array

cmd/wallet/main.go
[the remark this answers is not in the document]
esuwu: My point is that `show` plays the role of `showing an existing wallet` and I'll add description about it

cmd/wallet/main.go
```diff
@@ -1,163 +1,391 @@
...
+var examples = `
+Examples:
+	./wallet -show -wallet <wallet path> // Show existing wallet credentials
+	./wallet -new // Generate a seed phrase and a wallet 
+	./wallet -seed-phrase "..." // Import a seed phrase
+	./wallet -seed-phrase-base58 "..." // Import a Base58 encoded seed phrase
+	./wallet -account-seed "..." // Import a Base58 encoded account seed
 `
 
 type Opts struct {
```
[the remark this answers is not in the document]
esuwu: What's the difference between passing `PathToWallet` and passing a wallet file? We have pass it to functions anyway

alexeykiselev (approved):

PR #993, opened 2023-01-13
'Consensus and MetaMask updates' feature

nickeskov:


pkg/settings/feature_string.go
```diff
@@ -0,0 +1,42 @@
+// Code generated by "stringer -type=Feature"; DO NOT EDIT.
```
alexeykiselev, line 1: Why do we need `String`? No usages, internal name not equal to description.

pkg/settings/features.go
```diff
@@ -16,13 +19,18 @@ const (
...
-	InvokeExpression // 18
+	BlockReward           // 14
+	BlockV5               // 15
+	RideV5                // 16
+	RideV6                // 17
+	ConsensusImprovements // 18
+	InvokeExpression      // 19
 )
 
+func (f Feature) Description() string {
```
alexeykiselev, line 30: Why do we need `Description`. No usages again.

alexeykiselev: `Implemented` would be useful. But description only used in API in iteration over all features.
I think such functions should be introduced along with code that use them.

alexeykiselev (changes requested):

pkg/settings/feature_string.go
nickeskov: Fixed

pkg/settings/features.go
nickeskov: Agree, fixed

alexeykiselev (approved):

PR #996, opened 2023-01-17
Peers logs improvement

nickeskov:


pkg/node/peer_manager/peer_manager.go
```diff
@@ -109,27 +109,28 @@ func (a *PeerManagerImpl) NewConnection(p peer.Peer) error {
...
 		err := errors.Errorf(
-			"versions are too different, current %s, connected %s",
+			"versions are too different, current %s, connected %s (peer '%s')",
 			a.version.String(),
 			p.Handshake().Version.String(),
+			p.ID(),
 		)
 		a.restrict(p, now, err.Error())
 		_ = p.Close()
-		return proto.NewInfoMsg(err)
```
alexeykiselev, line 118: Why move this error logging from DEBUG to ERROR level?

pkg/node/peer_manager/peer_manager.go
```diff
@@ -109,27 +109,28 @@ func (a *PeerManagerImpl) NewConnection(p peer.Peer) error {
...
+		return err
 	}
 	if p.Handshake().AppName != a.networkName {
 		err := errors.Errorf("peer '%s' has the invalid network name '%s', required '%s'",
 			p.ID(), p.Handshake().AppName, a.networkName)
 		a.restrict(p, now, err.Error())
 		_ = p.Close()
-		return proto.NewInfoMsg(err)
```
alexeykiselev, line 125: And this one. Those errors are very common and they will bloat the log with error messages. Has meaning only on DEBUG level.

alexeykiselev (changes requested):

pkg/node/peer_manager/peer_manager.go
nickeskov: That shouldn't be there. I forgot to return it back. Fixed.

pkg/node/peer_manager/peer_manager.go
nickeskov: The same as above, fixed.

alexeykiselev (approved):

PR #1001, opened 2023-01-23
Added height and version to debug/statehash

esuwu:


esuwu:
https://github.com/wavesplatform/gowaves/pull/1003

PR #1003, opened 2023-01-24
Added new fields to debug/statehash/<height> API.           Added new method debug/statehash/last

esuwu:


pkg/proto/types.go
```diff
@@ -4216,3 +4202,31 @@ func (s *StateHash) UnmarshalJSON(value []byte) error {
...
+			DigestWrapped(s.SponsorshipHash),
+			DigestWrapped(s.AliasesHash),
+			DigestWrapped(s.WavesBalanceHash),
+			DigestWrapped(s.AssetBalanceHash),
+			DigestWrapped(s.LeaseBalanceHash),
+		},
+	}
+}
+
+type StateHashJSDebug struct {
```
nickeskov, line 4224: How can I extract `StateHash` struct from this type?

nickeskov (changes requested):

pkg/api/node_api.go
```diff
@@ -612,12 +612,13 @@ func (a *NodeApi) stateHash(w http.ResponseWriter, r *http.Request) error {
 		// TODO(nickeskov): which error it should send?
 		return &BadRequestError{err}
 	}
-
 	stateHash, err := a.state.StateHashAtHeight(height)
 	if err != nil {
 		return errors.Wrapf(err, "failed to get state hash at height %d", height)
 	}
-	if err := trySendJson(w, stateHash); err != nil {
+	stateHashDebug := proto.NewStateHashJSDebug(*stateHash, height, a.app.version().Version)
```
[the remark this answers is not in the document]
esuwu: Client library forwards requests to a given URL
What exactly do you think we should update?

pkg/api/node_api.go
```diff
@@ -605,19 +605,43 @@ func (a *NodeApi) nodeProcesses(w http.ResponseWriter, _ *http.Request) error {
...
 
-	stateHash, err := a.state.StateHashAtHeight(height)
+	if err := trySendJson(w, stateHashDebug); err != nil {
+		return errors.Wrap(err, "stateHash")
+	}
+	return nil
+}
+
+func (a *NodeApi) stateHashLast(w http.ResponseWriter, _ *http.Request) error {
+	height, err := a.state.Height()
```
alexeykiselev, line 636: Please reduce the height by 1 with corresponding checks. Scala node doesn't return state hash for the last block (liquid block as they call it) so they return last state hast at height-1.

pkg/api/node_api.go
```diff
@@ -605,19 +605,43 @@ func (a *NodeApi) nodeProcesses(w http.ResponseWriter, _ *http.Request) error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get the last height")
+	}
+	stateHashDebug, err := a.stateHashDebug(height)
 	if err != nil {
-		return errors.Wrapf(err, "failed to get state hash at height %d", height)
+		return errors.Wrap(err, "failed to get state hash debug")
```
alexeykiselev, line 642: Please remove error wrap here or replace message with more meaningful "failed to get last state hash".  From this message we can see at what API path error occured.

alexeykiselev (changes requested):

pkg/api/node_api.go
```diff
@@ -605,19 +605,43 @@ func (a *NodeApi) nodeProcesses(w http.ResponseWriter, _ *http.Request) error {
...
+
+func (a *NodeApi) stateHashLast(w http.ResponseWriter, _ *http.Request) error {
+	height, err := a.state.Height()
+	if err != nil {
+		return errors.Wrap(err, "failed to get the last height")
+	}
+	stateHashDebug, err := a.stateHashDebug(height - 1)
 	if err != nil {
-		return errors.Wrapf(err, "failed to get state hash at height %d", height)
+		return errors.Wrap(err, "failed to the get last state hash")
```
alexeykiselev, line 642: Message looks wrong :)

alexeykiselev (changes requested):

pkg/proto/types.go
```diff
@@ -4221,12 +4221,42 @@ func (s *StateHash) toStateHashJS() stateHashJS {
 	}
 }
 
-type StateHashJSDebug struct {
+type StateHashDebug struct {
 	stateHashJS
 	Height  uint64 `json:"height"`
```
alexeykiselev, line 4226: Consider adding `omitempty` to height and version fields. What if we request the previous version of node that doesn't support new extended state hash API?

alexeykiselev (changes requested):

nickeskov (approved):

alexeykiselev (approved):

PR #1004, opened 2023-01-24
Recipient struct refactoring

nickeskov:


pkg/proto/actions_validator.go
```diff
@@ -148,7 +148,7 @@ func ValidateAttachedPaymentScriptAction(action *AttachedPaymentScriptAction, re
 				return errors.Wrap(err, "failed to validate TransferScriptAction")
 			}
 		}
-		if action.Recipient.Address.Equal(senderAddress) {
+		if action.Recipient.Address().Equal(senderAddress) {
```
alexeykiselev, line 151: Isn't that dangerous? What if there is no address inside of recipient? The `nil` will be returned in this case that lead to panic on calling `Equal`. May be better add functions `EqualAddress` and `EqualAlias` to `Recipient` interface?

alexeykiselev: Or may be better to use `Recipient.Eq()` in such cases?

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #1014, opened 2023-02-06
add cli flags for blacklisting

artemreyt:


cmd/node/node.go
```diff
@@ -66,6 +66,8 @@ var (
 	enableMetaMaskAPI          = flag.Bool("enable-metamask", true, "Enables/disables metamask API.")
 	enableMetaMaskAPILog       = flag.Bool("enable-metamask-log", false, "Enables/disables metamask API logging.")
 	enableGrpcApi              = flag.Bool("enable-grpc-api", false, "Enables/disables gRPC API.")
+	enableBlacklisting         = flag.Bool("enable-blacklisting", true, "Enables or disables blacklisting of peers.")
+	blackListResidenceTime     = flag.Duration("black-list-residence-time", 15*time.Minute, "Period of time for which the information about external peer stays in the blacklist.")
```
alexeykiselev, line 70: Seems like `-disable-blacklisting` with default value set to `false` will be better. Flag `-enable-blacklisting` with default value set to `true` hard to use without passing a flag value if user want to disable it.
With introduction of `-blacklist-residence-time` we can use one flag to set interval and disable blacklisting. By passing zero value to `-blacklist-residence-time` we can disable blacklisting.
And why default interval is 15 minutes, is it default value from scala-node? We had 5 minutes.

alexeykiselev (changes requested):

cmd/node/node.go
nickeskov: And don't forget to add these flag(s) to `debugCommandLineParameters`.

artemreyt: > Seems like `-disable-blacklisting` with default value set to `false` will be better. Flag `-enable-blacklisting` with default value set to `true` hard to use without passing a flag value if user want to disable it.

flag removed.

> With introduction of `-blacklist-residence-time` we can use one flag to set interval and disable blacklisting. By passing zero value to `-blacklist-residence-time` we can disable blacklisting.

done

> And why default interval is 15 minutes, is it default value from scala-node? We had 5 minutes.

Yes, from scala-node.

artemreyt: > And don't forget to add these flag(s) to `debugCommandLineParameters`.

added.

alexeykiselev (approved):

PR #1019, opened 2023-02-08
Groth16 update library

esuwu:


alexeykiselev (approved):

alexeykiselev (dismissed):

pkg/crypto/internal/groth16/bls12_381/verifying_key.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func FromBytesToVerifyingKey(vkBytes []byte) (gnark.VerifyingKey, error) {
+	var bvk BellmanVerifyingKeyBl12381
+	_, err := bvk.ReadFrom(bytes.NewReader(vkBytes))
+	if err != nil {
+		return nil, err
+	}
+	var b bytes.Buffer
+	buf := bufio.NewWriter(&b)
```
nickeskov, line 96: I don't see here reasonable usage of this `buf`. You're using `Flush` later, but before I can't find any writes to `buf`.

pkg/crypto/internal/groth16/bn256/verifying_key.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+	_, err = bvk.ReadFrom(bytes.NewReader(vkBytes))
+	if err != nil {
+		return nil, err
+	}
+
+	var b bytes.Buffer
+	buf := bufio.NewWriter(&b)
```
nickeskov, line 102: I don't see here reasonable usage of this `buf`. You're using `Flush` later, but before I can't find any writes to `buf`.

nickeskov (changes requested):
I also noticed that `gnark` uses internal logger. I think it should be disabled in our package where we use `gnark`.

nickeskov (approved):

alexeykiselev (approved):

PR #1022, opened 2023-02-09
Added size check on marshaling of byte slice with size.

alexeykiselev:
Comments fixed.
Unused function removed.

### DIFF
--- a/pkg/proto/common.go
+++ b/pkg/proto/common.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/proto/common.go
```diff
@@ -76,11 +77,15 @@ func StringWithUInt32Len(buf []byte) (string, error) {
...
-	sl := uint16(len(data))
-	binary.BigEndian.PutUint16(buf, sl)
+// PutBytesWithUInt16Len prepends given buf with 2 bytes of its length.
+func PutBytesWithUInt16Len(buf []byte, data []byte) error {
+	l := len(data)
+	if l > math.MaxInt16 {
+		return errors.Errorf("invalid data size %d", l)
+	}
+	binary.BigEndian.PutUint16(buf, uint16(l))
 	copy(buf[2:], data)
```
nickeskov, line 87: I think `len(buf) >= 2 + len(data)` is necessary here.

pkg/proto/common.go
```diff
@@ -98,11 +103,15 @@ func BytesWithUInt16Len(buf []byte) ([]byte, error) {
...
-	sl := uint32(len(data))
-	binary.BigEndian.PutUint32(buf, sl)
+// PutBytesWithUInt32Len prepends given buf with 4 bytes of its length.
+func PutBytesWithUInt32Len(buf []byte, data []byte) error {
+	l := len(data)
+	if l > math.MaxInt32 {
+		return errors.Errorf("invalid data size %d", l)
+	}
+	binary.BigEndian.PutUint32(buf, uint32(l))
 	copy(buf[4:], data)
```
nickeskov, line 113: I think `len(buf) >= 4 + len(data)` is necessary here.

nickeskov (changes requested):

pkg/proto/common.go
alexeykiselev: Yes, you are right, I suppose those checks was left out because the earlier version of this function doesn't return an error.

alexeykiselev: added

pkg/proto/common.go
alexeykiselev: added

nickeskov (approved):

PR #1048, opened 2023-03-14
add wallet node alias api

artemreyt:


pkg/state/aliases.go
```diff
@@ -258,3 +258,35 @@ func (a *aliases) disabledAliases() (map[string]struct{}, error) {
...
+		}
+	}()
+
+	aliases := []string{}
+	for iter.Next() {
+		keyBytes := iter.Key()
+		recordBytes := iter.Value()
+		var record aliasRecord
+		if err := record.unmarshalBinary(recordBytes); err != nil {
+			return nil, errors.Errorf("failed to unmarshal record: %v", err)
```
nickeskov, line 281: errors.Wrap

pkg/state/aliases.go
```diff
@@ -258,3 +258,35 @@ func (a *aliases) disabledAliases() (map[string]struct{}, error) {
...
+		}
+		var key aliasKey
+		if err := key.unmarshal(keyBytes); err != nil {
+			return nil, err
```
nickeskov, line 285: I think additional info should be added to this error. In other words, wrapped

pkg/state/aliases.go
```diff
@@ -258,3 +258,35 @@ func (a *aliases) disabledAliases() (map[string]struct{}, error) {
...
+		}
+		if !record.info.stolen && record.info.addr.Equal(addr) {
+			aliases = append(aliases, key.alias)
+		}
```
nickeskov, line 289: I guess here you're scanning all aliases in the state db. It's not the best solution, definitely
You need to create a new relation between address and its alias in order to optimally return aliases by the given address
And don't forget to increment state db version (`pkg/state/constants.go`, `StateVersion`) because it's incompatible change (state should be re-imported)

pkg/state/state.go
```diff
@@ -1629,6 +1629,14 @@ func (s *stateManager) AddrByAlias(alias proto.Alias) (proto.WavesAddress, error
 	return *addr, nil
 }
 
+func (s *stateManager) AliasesByAddr(addr *proto.WavesAddress) ([]string, error) {
```
nickeskov, line 1632: Use just `proto.WavesAddress`, not `*proto.WavesAddress`. Pointer can be `nil`, struct can't.

pkg/state/aliases.go
```diff
@@ -258,3 +258,35 @@ func (a *aliases) disabledAliases() (map[string]struct{}, error) {
 	}
 	return als, nil
 }
+
+func (a *aliases) aliasesByAddr(addr *proto.WavesAddress) ([]string, error) {
```
nickeskov, line 262: Use just `proto.WavesAddress`, not `*proto.WavesAddress`. Pointer can be `nil`, struct can't.

nickeskov (changes requested):

pkg/api/errors/validation.go
```diff
@@ -168,3 +169,13 @@ func NewCustomValidationError(message string) *CustomValidationError {
 		},
 	}
 }
+
+func NewAliasDoesNotExistError(aliasFull string) *AliasDoesNotExistError {
+	return &AliasDoesNotExistError{
+		genericError: genericError{
+			ID:       AliasDoesNotExistErrorID,
+			HttpCode: http.StatusNotFound,
+			Message:  fmt.Sprintf("alias '%s' doesn't exist", aliasFull),
```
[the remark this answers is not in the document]
nickeskov: Yes, It matches.

pkg/api/node_api.go
```diff
@@ -478,6 +480,126 @@ func (a *NodeApi) PeersConnect(w http.ResponseWriter, r *http.Request) error {
...
+	type addrResponse struct {
+		Address string `json:"address"`
+	}
+
+	aliasShort := chi.URLParam(r, "alias")
+
+	alias := proto.NewAlias(a.app.scheme(), aliasShort)
+	if _, err := alias.Valid(a.app.scheme()); err != nil {
+		msg := err.Error()
+		return apiErrs.NewCustomValidationError(msg)
```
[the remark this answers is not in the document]
nickeskov: Quite tricky, but it's still ok

pkg/api/node_api.go
```diff
@@ -478,6 +480,126 @@ func (a *NodeApi) PeersConnect(w http.ResponseWriter, r *http.Request) error {
...
+
+	addr, err := proto.NewAddressFromString(addrBase58)
+	if err != nil {
+		return &apiErrs.InvalidAddressError{}
+	}
+
+	aliases, err := a.app.AliasesByAddr(addr)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find aliases by addr")
+	}
```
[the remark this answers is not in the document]
nickeskov: UPD: It'll return empty array, not `nil`.

pkg/api/node_api.go
```diff
@@ -478,6 +479,130 @@ func (a *NodeApi) PeersConnect(w http.ResponseWriter, r *http.Request) error {
...
+		return errors.Wrapf(err, "failed to find process (self) with pid %d", selfPid)
+	}
+
+	interrupt := os.Interrupt
+	if err := p.Signal(interrupt); err != nil {
+		return errors.Wrapf(err,
+			"failed to send signal %q to self process with pid %d", interrupt, selfPid)
+	}
+
+	if err := trySendJson(w, resp{Stopped: true}); err != nil {
```
alexeykiselev, line 582: The chances of sending the response are not great. Should we consider not implementing this method at all?

pkg/state/history_storage.go
```diff
@@ -15,7 +15,7 @@ var errEmptyHist = errors.New("empty history for this record")
 type blockchainEntity byte
 
 const (
-	alias blockchainEntity = iota
+	alias blockchainEntity = iota + 1
```
alexeykiselev, line 18: What is the point of this change?

pkg/state/keys.go
```diff
@@ -28,7 +28,7 @@ const (
 // Primary prefixes for storage keys
 const (
 	// Balances.
-	wavesBalanceKeyPrefix byte = iota
+	wavesBalanceKeyPrefix byte = iota + 1
```
alexeykiselev, line 31: Why we should not use zero as prefix start? Don't get it.

alexeykiselev (changes requested):

pkg/api/node_api.go
nickeskov: Removed

pkg/state/aliases.go
```diff
@@ -65,51 +67,154 @@ func (r *aliasRecord) unmarshalBinary(data []byte) error {
...
+}
+
+func (a *aliases) addOrUpdateAddressToAliasesRecord(aliasStr string, addr proto.WavesAddress, blockID proto.BlockID) error {
+	var (
+		key         = addressToAliasesKey{addressID: addr.ID()}
+		recordBytes []byte
+		err         error
+	)
+	keyBytes := key.bytes()
+	// TODO: is it ok to use newestTopEntryData instead of topEntryData?
```
alexeykiselev, line 202: Yes, it's correct to use `newest` function, because in this case you will see the aliases created by other transactions in the same block or the bunch of blocks.

pkg/state/aliases.go
```diff
@@ -65,51 +67,154 @@ func (r *aliasRecord) unmarshalBinary(data []byte) error {
...
+type addressToAliasesRecord struct {
+	aliases []string
+}
+
+func (r *addressToAliasesRecord) marshalBinary() ([]byte, error) {
+	var (
+		b   []byte
+		err error
+	)
+	for _, a := range r.aliases {
```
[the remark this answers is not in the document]
nickeskov: And I've compared binary and CBOR serialized data and their size has been almost the same

alexeykiselev (approved):

alexeykiselev (dismissed):

alexeykiselev (approved):

alexeykiselev (approved):

nickeskov (approved):

PR #1068, opened 2023-04-11
HTTP API '/assets/details'

nickeskov:


pkg/proto/info.go
```diff
@@ -63,14 +63,16 @@ func (i *FullAssetInfo) ToProtobuf(scheme Scheme) (*g.AssetInfoResponse, error)
...
+	ID                crypto.Digest
+	Quantity          uint64
+	Decimals          byte
+	Issuer            WavesAddress
+	IssuerPublicKey   crypto.PublicKey
+	Reissuable        bool
+	Scripted          bool
+	Sponsored         bool
+	IssueHeight       Height
+	IssueTxPosInBlock uint64 // TODO: size in bits can be reduced
```
alexeykiselev, line 75: Really, why `uint64`? `uint32` would be enough as in scala implementation

pkg/state/appender.go
```diff
@@ -628,6 +631,7 @@ func (a *txAppender) appendBlock(params *appendBlockParams) error {
 			consensusImprovementsActivated: consensusImprovementsActivated,
 			invokeExpressionActivated:      invokeExpressionActivated,
 			validatingUtx:                  false,
+			txPosInBlock:                   uint64(i + 1),
```
alexeykiselev, line 634: This assumption is incorrect. The `txPosInBlock` field indicates the issuance position of an asset within a block. It can be incremented by either the `IssueTransaction` or the `IssueAction` of an `InvokeTransaction`. Perhaps the field's name does not accurately reflect its meaning. However, it should be noted that other transactions and actions do not modify this value.

alexeykiselev (changes requested):

pkg/proto/info.go
nickeskov: done

pkg/ride/environment.go
```diff
@@ -20,6 +20,8 @@ type WrappedState struct {
 	diff                      diffState
 	cle                       rideAddress
 	scheme                    proto.Scheme
+	height                    proto.Height
+	txPosInBlock              uint64
```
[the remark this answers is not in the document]
nickeskov: done

alexeykiselev (dismissed):

Anton-Rampage (approved):

alexeykiselev (approved):

PR #1071, opened 2023-04-12
Add any and throw types, fix some test, add tests scripts

Anton-Rampage:


pkg/ride/compiler/ast_parser_test.go
```diff
@@ -868,23 +869,23 @@ func cursed() = [][0]
...
-//			code, err := embedScripts.ReadFile("test_scripts/" + file.Name())
-//			require.NoError(t, err)
-//			res, _, err := cli.Utils.ScriptCompile(context.Background(), string(code))
-//			require.NoError(t, err)
-//			compareScriptsOrError(t, string(code), false, strings.TrimPrefix(res.Script, "base64:"))
-//		}
-//	}
+//go:embed test_scripts
+var embedScripts embed.FS
+
```
alexeykiselev, line 874: Shouldn't we execute tests that require communication with an external resource (node) only when a special option is provided?"
```suggestion
var integration bool

func init() {
	flag.BoolVar(&integration, "integration", false, "Execute integration tests. False by default.")
}

func TestBigScripts(t *testing.T) {
	if !integration {
		t.Skip("Skipping integration test")
	}

```
To run the test the `-integration` flag should be provided:
```bash
go test -integration ./ast_parser_test.go
```

alexeykiselev (changes requested):

alexeykiselev (approved):

nickeskov (approved):

PR #1079, opened 2023-04-20
Add remove unused code in compiler

Anton-Rampage:


pkg/ride/compiler/compaction.go
```diff
@@ -193,3 +193,103 @@ func (c *Compaction) processExpr(node ast.Node) ast.Node {
...
+		if _, exists := map2[key]; !exists {
+			difference[key] = value
+		}
+	}
+
+	return difference
+}
+
+func appendMapNames(to, from map[string]struct{}) map[string]struct{} {
+	res := make(map[string]struct{})
```
alexeykiselev, line 234: We can improve by creating new map with a size of `to` and `from` maps combined.
Also, there's a function experimental standard function https://pkg.go.dev/golang.org/x/exp/maps#Copy that makes almost what you need. But it copies keys and values from source to destination maps and doesn't return any map. May be not so useful in this case.

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #1111, opened 2023-06-01
Fixed behaviour of Ride function blockInfoByHeight for non-positive h…

alexeykiselev:
…eight values.


pkg/ride/functions_proto.go
```diff
@@ -836,9 +836,12 @@ func blockInfoByHeight(env environment, args ...rideType) (rideType, error) {
 		return nil, errors.Wrap(err, "blockInfoByHeight")
 	}
 	height := proto.Height(i)
+	if height <= 0 {
+		return rideUnit{}, nil
+	}
```
nickeskov, line 841: Maybe add test for this?

pkg/ride/functions_proto.go
```diff
@@ -836,9 +836,12 @@ func blockInfoByHeight(env environment, args ...rideType) (rideType, error) {
...
 	header, err := env.state().NewestHeaderByHeight(height)
 	if err != nil {
-		return nil, errors.Wrap(err, "blockInfoByHeight")
+		return rideUnit{}, nil
```
nickeskov, line 844: Is it necessary to ignore the error here?

nickeskov (changes requested):

pkg/ride/functions_proto_test.go
```diff
@@ -736,7 +737,30 @@ func TestAssetInfoV4(t *testing.T) {
...
+	for _, test := range []struct {
+		te   *testEnv
+		args []rideType
+		fail bool
+		r    rideType
+	}{
+		{env, []rideType{rideInt(0)}, false, rideUnit{}},
+		{env, []rideType{rideInt(-1)}, false, rideUnit{}},
+		{env, []rideType{rideInt(2)}, false, obj},
+		{env, []rideType{rideInt(3)}, true, nil},
```
nickeskov, line 754: @alexeykiselev Is it valid that error will be returned instead of unit for blocks in **future** (non non existent blocks)?

pkg/state/script_caller.go
```diff
@@ -187,7 +189,9 @@ func (a *scriptCaller) callAssetScriptCommon(env *ride.EvaluationEnvironment, se
 		}
 		env.SetThisFromFullAssetInfo(assetInfo)
 	}
-	env.SetLastBlock(params.blockInfo)
+	if err := env.SetLastBlockFromBlockInfo(params.blockInfo); err != nil {
+		return nil, err
```
nickeskov, line 193: Maybe add info for the error?

nickeskov (changes requested):

pkg/ride/functions_proto_test.go
alexeykiselev: No, it won't, it is just a mock behaviour to test the error handling. In real life in this case the errNotFound will be returned and handled by returning Unit from function.

pkg/state/script_caller.go
alexeykiselev: No error wrapping before the script execution in this function, so it's the same.

PR #1113, opened 2023-06-06
Update asset info test suite

irina-pereiaslavskaia:
New tests for Update Asset Info Tx


alexeykiselev (dismissed):

itests/update_asset_info_api_test.go
```diff
@@ -0,0 +1,317 @@
...
+		for _, iv := range issue_versions {
+			reissuable := testdata.GetCommonIssueData(&suite.BaseSuite).Reissuable
+			itx := issue_utilities.IssueBroadcastWithTestData(&suite.BaseSuite, reissuable, iv, waitForTx)
+			tdmatrix := testdata.GetUpdateAssetInfoPositiveDataMatrix(&suite.BaseSuite, itx.TxID)
+			for name, td := range tdmatrix {
+				height := utl.GetHeight(&suite.BaseSuite)
+				caseName := utl.GetTestcaseNameWithVersion(name, v)
+				suite.Run(caseName, func() {
+					//***wait n blocks***
+					utl.WaitForHeight(&suite.BaseSuite, height+1)
```
nickeskov, line 34: This would work, definitely. But you can do better.
You can issue assets before all testcases in amount equal of number of  testcases in the test.
If you use different asset for each testcase, you you'll need wait only once after assets issues.

itests/update_asset_info_api_test.go
```diff
@@ -0,0 +1,317 @@
...
+	issue_versions := issue_utilities.GetVersions(&suite.BaseSuite)
+	waitForTx := true
+	for _, v := range versions {
+		for _, iv := range issue_versions {
+			reissuable := testdata.GetCommonIssueData(&suite.BaseSuite).Reissuable
+			itx := issue_utilities.IssueBroadcastWithTestData(&suite.BaseSuite, reissuable, iv, waitForTx)
+			tdmatrix := testdata.GetUpdateAssetInfoPositiveDataMatrix(&suite.BaseSuite, itx.TxID)
+			for name, td := range tdmatrix {
+				height := utl.GetHeight(&suite.BaseSuite)
+				caseName := utl.GetTestcaseNameWithVersion(name, v)
```
nickeskov, line 31: I think `iv` should be also included in the testcase name.

nickeskov (changes requested):
I don't like to repeat the same issues multiple times, so I've written them in one place. These issues can be applied for similar places in the code of current PR.

itests/update_asset_info_api_test.go
irina-pereiaslavskaia: Done

itests/update_asset_info_api_test.go
irina-pereiaslavskaia: Fix

DrBlast (approved):

nickeskov (approved):

PR #1115, opened 2023-06-08
Fix error with state hash cmp

irina-pereiaslavskaia:
Fix teardown method for test suites

### DIFF
--- a/itests/node_client/node_client.go
+++ b/itests/node_client/node_client.go
@@ -7,7 +7,6 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
-
 	d "github.com/wavesplatform/gowaves/itests/docker"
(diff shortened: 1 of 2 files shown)


itests/node_client/node_client.go
```diff
@@ -85,6 +84,17 @@ func (c *NodesClients) WaitForHeight(t *testing.T, height uint64) uint64 {
 	return hs
 }
 
+func (c *NodesClients) WaitForStateHashEquality(t *testing.T) {
```
alexeykiselev, line 87:
```suggestion
func (c *NodesClients) WaitForStateHashEquality(t *testing.T) {
    equal := false
    for i := 0; i < 3; i++ {
        h := c.WaitForNewHeight(t)
        if c.StateHashCmp(h) {
           equal := true
           break
       }
    }
    assert.True(equal)
}
```

irina-pereiaslavskaia: Done

alexeykiselev (approved):

PR #1163, opened 2023-07-19
Added build of ARM64 for gowaves-testnet DEB package.

alexeykiselev:


Makefile
```diff
@@ -12,25 +12,13 @@ export GO111MODULE=on
 
 .PHONY: vendor vetcheck fmtcheck clean build gotest update-go-deps
 
-all: vendor vetcheck fmtcheck build gotest mod-clean
+all: vendor vetcheck fmtcheck gotest mod-clean
```
nickeskov, line 15: I think `release-node` should be included to this target.

nickeskov: Or even better: `build-node-native` target.

nickeskov (approved):

PR #1166, opened 2023-07-21
Add rewards api methods

Anton-Rampage:


pkg/state/state.go
```diff
@@ -1527,6 +1581,10 @@ func (s *stateManager) addBlocks() (*proto.Block, error) {
 	if err := s.stor.handleStateHashes(height, ids); err != nil {
 		return nil, wrapErr(ModificationError, err)
 	}
+	// Store total amount and rewards for each of new blocks.
+	if err = s.stor.handleRewardsAndTotalWavesAmount(height, ids, rewards); err != nil {
+		return nil, wrapErr(ModificationError, err)
+	}
```
nickeskov, line 1587: I think this data should be saved in the main cycle above, i.e. in the process of block application.

nickeskov (changes requested):
Additional issues:
- [x] Constant `StateVersion` in `pkg/state/constants.go` should be incremented because compatibility with previous state version has been lost.
- [ ] `rewardCalculator.performCalculation` also should use new `stateManager`'s or internal micro-state's methods for historical data. Also note, that rewards data by current height **must** exist at the moment of block applying.
- [x] Pay attention to `stateManager.blockRewards` method. Also should be changed.

If you need additional info feel free to ask.

pkg/state/history_storage.go
```diff
@@ -194,6 +196,16 @@ var properties = map[blockchainEntity]blockchainEntityProperties{
 		needToCut:    true,
 		fixedSize:    false,
 	},
+	blockRewardAtHeight: {
+		needToFilter: true,
+		needToCut:    true,
+		fixedSize:    false,
```
alexeykiselev, line 202: We can set `needToCut` to false to store all historical values of `blockRewards` and `totalWavesAmount`. For example look at `featureVote`.

pkg/state/state.go
```diff
@@ -1527,6 +1581,10 @@ func (s *stateManager) addBlocks() (*proto.Block, error) {
 	if err := s.stor.handleStateHashes(height, ids); err != nil {
 		return nil, wrapErr(ModificationError, err)
 	}
+	// Store total amount and rewards for each of new blocks.
```
alexeykiselev, line 1584: Initially I thought that we won't save reward and totalWavesAmount for every block. It's a huge waste of space. For the first 1_740_000 blocks on MainNet it's only zeroes. And after that we don't have to store it for every block We have to store new value of reward only if it's changed. From 0 to 6 on activation of Feature 14 and so on. For now it will be only one value. 
The total Waves amount can be calculated.

alexeykiselev (changes requested):

alexeykiselev (approved):

pkg/state/monetary_policy.go
```diff
@@ -238,12 +246,12 @@ func (m *monetaryPolicy) totalAmountAtHeight(
 			continue
 		}
 		if height > change.Height && !isNotLast {
-			curTotalAmount += change.Reward * (height - change.Height)
+			curTotalAmount += change.Reward * (height - (change.Height - 1))
```
nickeskov, line 249: It's nice to have `(change.Height - 1)` as a variable.

pkg/state/monetary_policy_test.go
```diff
@@ -165,6 +165,51 @@ func TestFinishRewardVoting(t *testing.T) {
 	}
 }
 
+func TestRewardAtHeight(t *testing.T) {
```
nickeskov, line 168: Refactor to table test

pkg/state/monetary_policy_test.go
```diff
@@ -165,6 +165,51 @@ func TestFinishRewardVoting(t *testing.T) {
...
+	reward, err := mo.rewardAtHeight(4, blockRewardActivationHeight)
+	require.NoError(t, err)
+	assert.Equal(t, initialReward, reward)
+
+	reward, err = mo.rewardAtHeight(10, blockRewardActivationHeight)
+	require.NoError(t, err)
+	assert.Equal(t, initialReward+100000000, reward)
+}
+
+func TestTotalWavesAmountAtHeight(t *testing.T) {
```
nickeskov, line 191: Refactor to table test

nickeskov (changes requested):

nickeskov (changes requested):

pkg/state/monetary_policy.go
```diff
@@ -56,52 +43,29 @@ func (r *rewardVotesRecord) unmarshalBinary(data []byte) error {
...
+	return &monetaryPolicy{hs: hs, settings: settings}
 }
 
 func (m *monetaryPolicy) reward() (uint64, error) {
-	var record blockRewardRecord
-	b, err := m.hs.newestTopEntryData(blockRewardKeyBytes)
-	if isNotFoundInHistoryOrDBErr(err) {
-		return m.settings.InitialBlockReward(), nil
+	rewardsChanges, err := m.getRewardChanges()
+	if isNotFoundInHistoryOrDBErr(err) || len(rewardsChanges) == 0 {
```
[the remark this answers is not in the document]
alexeykiselev: Fixed like this:
```go
	rewardsChanges, err := m.getRewardChanges()
	if err != nil {
		if isNotFoundInHistoryOrDBErr(err) {
			return m.settings.InitialBlockReward, nil
		}
		return 0, err
	}
	if len(rewardsChanges) == 0 {
		return m.settings.InitialBlockReward, nil
	}
	return rewardsChanges[len(rewardsChanges)-1].Reward, nil
```

alexeykiselev (approved):

pkg/state/monetary_policy.go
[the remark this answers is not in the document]
nickeskov: It's even better

PR #1179, opened 2023-08-09
Implement rollback api method

Anton-Rampage:


pkg/api/node_api.go
```diff
@@ -649,31 +649,50 @@ func (a *NodeApi) unconfirmedSize(w http.ResponseWriter, _ *http.Request) error
...
+	type rollbackRequest struct {
+		Height                  uint64 `json:"rollbackTo"`
+		ReturnTransactionsToUtx bool   `json:"returnTransactionsToUtx"`
+	}
+
+	rollbackReq := &rollbackRequest{}
+	if err := tryParseJson(r.Body, rollbackReq); err != nil {
+		return errors.Wrap(err, "failed to parse RollbackToHeight body as JSON")
+	}
+	err := a.state.RollbackToHeight(rollbackReq.Height)
```
alexeykiselev, line 666: Consider error handling like at `pkg/api/node_api.go:95`. Function `state.RollbackToHeight` can return `RetrievalError` which probably should be considered as 404 HTTP error.

pkg/api/node_api.go
```diff
@@ -649,31 +649,50 @@ func (a *NodeApi) unconfirmedSize(w http.ResponseWriter, _ *http.Request) error
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to rollback to height %d", rollbackReq.Height)
+	}
+	block, err := a.state.BlockByHeight(rollbackReq.Height)
```
alexeykiselev, line 670: Consider returning 404 as above on a retrieval error.

pkg/api/node_api.go
alexeykiselev: Note that `RollbackTo` function doesn't return RetrievalError.

alexeykiselev (changes requested):

PR #1190, opened 2023-08-22
Logged first block of fail

esuwu:


esuwu:
I was using t.Log in this PR, but if there's something else I should use, let me know!

pkg/proto/types.go
```diff
@@ -4033,6 +4033,22 @@ type fieldsHashesJS struct {
 	LeaseBalanceHash  DigestWrapped `json:"leaseBalanceHash"`
 }
 
+func (fh FieldsHashes) EqualWith(other FieldsHashes) bool {
```
alexeykiselev, line 4036: Looks complicated. Why not just compare all fields one by one? And I think you can compare `Digest`s without using `bytes.Equal` because they are fixed arrays.

alexeykiselev (changes requested):

pkg/proto/types.go
```diff
@@ -4033,6 +4033,37 @@ type fieldsHashesJS struct {
 	LeaseBalanceHash  DigestWrapped `json:"leaseBalanceHash"`
 }
 
+func (s FieldsHashes) EqualWith(other FieldsHashes) bool {
```
alexeykiselev, line 4036: It can be even simpler, like:
```go
return s.DateEntryHash == other.DataEntryHash && s.AccountScriptHash == other.AccountScriptHash && ...
```
And name of the function better change to `Equal`.

alexeykiselev: Usually argument `other` named as `o`.

alexeykiselev (changes requested):

pkg/proto/types.go
esuwu: I would leave `other` as it is if that's okay. I don't like one-letter variables, they always make code hard to read.

nickeskov: @esuwu First issue is still relevant.

alexeykiselev (approved):

nickeskov (approved):

PR #1210, opened 2023-09-27
Add complexity to snapshots

esuwu:


esuwu:
This PR shouldn't be merged just yet

pkg/state/snapshot_generator_internal_test.go
```diff
@@ -1048,18 +1130,245 @@ func TestDefaultInvokeScriptSnapshot(t *testing.T) {
...
 	}
 
+	var expectedDataEntrySnapshot proto.DataEntriesSnapshot
+	idxExpectedDataSnapshot := 0
+	for idx, atomicSnapshot := range expectedSnapshot {
+		if dataEntryS, ok := atomicSnapshot.(*proto.DataEntriesSnapshot); ok {
+			idxExpectedDataSnapshot = idx
+			expectedDataEntrySnapshot = *dataEntryS
+		}
+	}
```
nickeskov, line 1340: Maybe move to a function?

pkg/state/snapshot_generator_internal_test.go
```diff
@@ -1048,18 +1130,245 @@ func TestDefaultInvokeScriptSnapshot(t *testing.T) {
...
+			}
+	*/
+	script := "AAIFAAAAAAAAAAQIAhIAAAAAAAAAAAEAAAABaQEAAAAEY2Fsb" +
+		"AAAAAAJAARMAAAAAgkBAAAADEJvb2xlYW5FbnRyeQAAAAICAAAABGJvb2wGCQAE" +
+		"TAAAAAIJAQAAAAxJbnRlZ2VyRW50cnkAAAACAgAAAANpbnQAAAAAAAAAAAEJAARMAAAA" +
+		"AgkBAAAADEludGVnZXJFbnRyeQAAAAICAAAAA2ludAAAAAAAAAAAAQkABEwAAAACCQEAAA" +
+		"AMSW50ZWdlckVudHJ5AAAAAgIAAAAEaW50MgAAAAAAAAAAAgkABEwAAAACCQEAAAALU3RyaW5nRW" +
+		"50cnkAAAACAgAAAANzdHICAAAAATEJAARMAAAAAgkBAAAAC1N0cmluZ0VudHJ5AAAAAgIAAAADc3RyAgAAAA" +
+		"ExCQAETAAAAAIJAQAAAAtTdHJpbmdFbnRyeQAAAAICAAAABHN0cjICAAAAATIFAAAAA25pbAAAAACkN9Gf"
+	scriptsBytes, err := base64.StdEncoding.DecodeString(script)
```
nickeskov, line 1236: We don't have to store compiled code here. Just use the compiler package and it will create parsed tree from ride script source code.

pkg/state/snapshot_generator.go
```diff
@@ -791,6 +867,18 @@ func (sg *snapshotGenerator) generateInvokeSnapshot(
 		snapshot = append(snapshot, &assetBalancesSnapshot[i])
 	}
 
+	if sg.IsFullNodeMode && scriptEstimation != nil && scriptRecipient != nil {
+		if scriptEstimation.isPresent() {
```
nickeskov, line 871: You don't have to check `scriptEstimation != nil`. It's equivalent to the `scriptEstimation.isPresent` implementation.
In which case `scriptRecipient` can be `nil`? `InvokeExpression` transaction?

pkg/state/snapshot_generator.go
```diff
@@ -745,8 +808,19 @@ func (sg *snapshotGenerator) atomicSnapshotsFromScriptActions(
...
-		dataEntrySnapshot := &proto.DataEntriesSnapshot{Address: address, DataEntries: entries}
+	senderDataEntries := make(map[proto.WavesAddress]proto.DataEntries)
+	for _, entry := range dataEntriesMap {
+		if _, ok := senderDataEntries[entry.addr]; ok {
+			entries := senderDataEntries[entry.addr]
+			entries = append(entries, entry.dataEntry)
+			senderDataEntries[entry.addr] = entries
+		} else {
+			senderDataEntries[entry.addr] = proto.DataEntries{entry.dataEntry}
+		}
```
nickeskov, line 819: `append` works fine with `nil` slices.

nickeskov (changes requested):

esuwu:
Import is ok on this commit [c10178b](https://github.com/wavesplatform/gowaves/pull/1210/commits/c10178b25cb19ebe72fdefdb9493cd8afe2ab56f)

pkg/state/snapshot_generator.go
esuwu: it's a map, not a slice

PR #1140, opened 2023-06-23
Add attachment field in order

Anton-Rampage:

PR #1255, opened 2023-11-30
Ride V8 library version added. Tests updated.

alexeykiselev:


pkg/ride/compiler/ast_parser.go
```diff
@@ -136,7 +136,7 @@ func (p *astParser) loadBuildInVarsToStackByVersion() {
 			Type: txType,
 		})
 	}
-	if p.tree.LibVersion >= ast.LibV4 && p.tree.LibVersion <= ast.LibV7 {
+	if p.tree.LibVersion >= ast.LibV4 && p.tree.LibVersion <= ast.LibV8 {
```
nickeskov, line 139: Maybe use `CurrentMaxLibraryVersion` instead of `ast.LibV8`?

pkg/ride/compiler/stdlib/types.go
```diff
@@ -615,6 +616,9 @@ func mustLoadDefaultTypes() map[ast.LibraryVersion]map[string]Type {
 		ast.LibV7: {
 			"Transaction": UnionType{Types: []Type{}},
 		},
+		ast.LibV8: {
+			"Transaction": UnionType{Types: []Type{}},
+		},
```
nickeskov, line 621: Consider using `for` cycle for initializing the map

pkg/ride/generate/internal/functions_generation.go
```diff
@@ -919,12 +931,24 @@ func evaluationCatalogueV7EvaluatorV1() map[string]int {
 	return m
 }
 
+func evaluationCatalogueV8EvaluatorV1() map[string]int {
```
nickeskov, line 934: Please, remind me, why we need `evaluationCatalogueEvaluatorV1` for all versions of ride since RideV6 included?

I can see the code inside `newEvaluationScope`:
```go
	ev := 1
	if env.rideV6Activated() {
		ev = 2
	}
	costs, err := selectEvaluationCostsProvider(v, ev)
	if err != nil {
		return evaluationScope{}, err
	}
```
Inside `selectEvaluationCostsProvider`:
```go
	switch v {
	case ast.LibV1, ast.LibV2:
		switch ev {
		case 1:
			return EvaluationCatalogueV2EvaluatorV1, nil
		default:
			return EvaluationCatalogueV2EvaluatorV2, nil
		}
		case ...
```
It means that we need catalogueV1 for evaluator V1. But `RideV6` must be activated for scripts V6 and above.
And `EvaluationCatalogueV6EvaluatorV1`, `EvaluationCatalogueV7EvaluatorV1` and `EvaluationCatalogueV8EvaluatorV1` are used nowhere.

pkg/ride/tree_estimatorV1.go
```diff
@@ -103,6 +104,8 @@ func newTreeEstimatorV1(tree *ast.Tree) (*treeEstimatorV1, error) {
 		r.scope = newEstimationScopeV1(ConstantsV6, CatalogueV6)
 	case ast.LibV7:
 		r.scope = newEstimationScopeV1(ConstantsV7, CatalogueV7)
+	case ast.LibV8:
+		r.scope = newEstimationScopeV1(ConstantsV8, CatalogueV8)
```
nickeskov, line 108: Why we need to handle RideV8 scripts for the first version of estimator when version 4 will be used for RideV8 scripts? Can be applied for other estimators (excluding version 4).

If this is intended logic, please, explain why

nickeskov (changes requested):

pkg/ride/compiler/ast_parser.go
alexeykiselev: It fixed in #1140.

pkg/ride/generate/internal/functions_generation.go
alexeykiselev: Yes, exactly. It was all made to appear similar just to avoid raising concerns. :)

pkg/ride/tree_estimatorV1.go
alexeykiselev: The estimation logic made independent of features activation. That's it.

nickeskov (approved):

PR #1257, opened 2023-11-30
Ride function replaceByIndex implemented and tested.

alexeykiselev:


pkg/ride/functions_list.go
```diff
@@ -86,6 +93,37 @@ func listAndElementArgs(args []rideType) (rideList, rideType, error) {
...
+	}
+	if args[0] == nil {
+		return nil, 0, nil, errors.Errorf("argument 1 is empty")
+	}
+	if args[1] == nil {
+		return nil, 0, nil, errors.Errorf("argument 2 is empty")
+	}
+	if args[2] == nil {
+		return nil, 0, nil, errors.Errorf("argument 3 is empty")
+	}
```
nickeskov, line 108: Can be replaced by `checkArgs` call.

pkg/ride/functions_list.go
```diff
@@ -4,10 +4,14 @@ import (
...
 
 	"github.com/pkg/errors"
+
 	"github.com/wavesplatform/gowaves/pkg/ride/math"
 )
 
-const maxListSize = 1000
+const (
+	maxListSize = 1000
+	threeArgs   = 3
```
nickeskov, line 13: I think `threeArgs` constant should be inside the function body where it is used.

nickeskov (changes requested):

nickeskov (approved):

PR #1264, opened 2023-12-07
Tests on correct scopes handling by estimator and evaluator added.

alexeykiselev:
Few compilation errors of FOLD fixed.
Comparison of Any type to other types fixed in compilation.


pkg/ride/compiler/stdlib/types.go
```diff
@@ -218,7 +218,7 @@ func (t SimpleType) Equal(other Type) bool {
 	if o, ok := other.(SimpleType); ok {
 		return t.Type == o.Type
 	}
-	return false
+	return other == AnyType
```
nickeskov, line 221: I think the correct comparison is `other.Equal(AnyType)` or `AnyType.Equal(other)` if other can be `nil`.
Though I don't understand why this comparison is necessary?

nickeskov (changes requested):

pkg/ride/compiler/stdlib/types.go
alexeykiselev: No, this is the `Equal` of SimpleType itself. Can't use it here.
Any should be equal to any other type. It fires when we create an empty list using `let l = []`. In this case the type of list is not defined and set to `Any`. Later then we check the correctness of `FOLD` we have to check that type of list equals to the type the accumulator. Whiteout this equality the FOLD rises the error, but it shouldn't. 
I'll consider using `AnyType.Equal` here.

nickeskov (approved):

PR #1307, opened 2024-01-29
Added new way to configure preactivation of features at any height.

alexeykiselev:
Integration tests moved to use a new way of preactivation with height.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -66,30 +66,30 @@ type blockchainEntitiesStorage struct {
 }
 
 func newBlockchainEntitiesStorage(hs *historyStorage, sets *settings.BlockchainSettings, rw *blockReadWriter, calcHashes bool) (*blockchainEntitiesStorage, error) {
-	assets := newAssets(hs.db, hs.dbBatch, hs)
-	balances, err := newBalances(hs.db, hs, assets, sets.AddressSchemeCharacter, calcHashes)
+	asts := newAssets(hs.db, hs.dbBatch, hs)
+	bss, err := newBalances(hs.db, hs, asts, sets.AddressSchemeCharacter, calcHashes)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/state/state.go
```diff
@@ -636,6 +638,16 @@ func (s *stateManager) applyPreActivatedFeatures(features []int16, blockID proto
 			return err
 		}
 	}
+	for _, fah := range featuresActivationHeights {
+		approvalRequest := &approvedFeaturesRecord{approvalHeight: fah.Height}
+		if err := s.stor.features.approveFeature(fah.ID, approvalRequest, blockID); err != nil {
+			return err
+		}
+		activationRequest := &activatedFeaturesRecord{activationHeight: fah.Height}
+		if err := s.stor.features.activateFeature(fah.ID, activationRequest, blockID); err != nil {
```
nickeskov, line 647: This method adds an `activation` record. But it's incorrect, because `newestIsActivated` will return `true` even if the blockchain height is lower than activation height.
The same can be applied for the approval stage and `newestIsApproved` method.

I think the good decision is to include activation hooks inside `stateManager.blockchainHeightAction`.

nickeskov (changes requested):

pkg/state/state.go
nickeskov: See #1308

alexeykiselev: Nope, it's totally correct. The idea of preactivation is to activate some features at genesis block prior everything else. I agree that activation of feature at genesis block but at certain height looks weird, but it was made for testing purposes only and must be used only for this. "Our" way of preactivation of features in genesis block and at genesis block height is the most correct way of preactivation and this way must be used for custom blockchains. That's why I save it for backward compatibility.

alexeykiselev: I simplified the configuration structure. Removed the unnecessary new structure and made it looks like in scala implementation.

PR #1332, opened 2024-02-13
Start applying from file since height 2 in importer.

nickeskov:


nickeskov:
This PR is incorrect. `nBlocks` parameter equals number of blocks to import WITHOUT genesis block.
Due to this misunderstanding the PR was created.
Closing as irrelevant.

PR #1333, opened 2024-02-13
Fix 'isNFT' flag initialization for assets

nickeskov:
This PR **breaks compatibility** with the previous state version.

### DIFF
--- a/pkg/state/assets.go
+++ b/pkg/state/assets.go
@@ -24,17 +24,19 @@ type assetInfo struct {
 	assetChangeableInfo
 }
 
+// initIsNFTFlag initializes IsNFT flag of assetConstInfo.
+// IsNFT flag must be initialized only once at asset issuing moment by calling this method.
 func (ai *assetInfo) initIsNFTFlag(fs featuresState) error {
 	if !ai.quantity.IsInt64() {
-		ai.isNFT = false
+		ai.IsNFT = false
 		return nil
 	}
-	ap := assetParams{ai.quantity.Int64(), int32(ai.decimals), ai.reissuable}
+	ap := assetParams{ai.quantity.Int64(), int32(ai.Decimals), ai.reissuable}
 	nft, err := isNFT(fs, ap)
 	if err != nil {
 		return err
 	}
-	ai.isNFT = nft
+	ai.IsNFT = nft
 	return nil
 }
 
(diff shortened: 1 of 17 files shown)


pkg/state/assets.go
```diff
@@ -23,70 +24,72 @@ type assetInfo struct {
...
 type assetConstInfo struct {
 	tail                 [proto.AssetIDTailSize]byte
 	issuer               crypto.PublicKey
 	decimals             uint8
 	issueHeight          proto.Height
+	isNFT                bool // isNFT flag must be initialized only once by calling assetInfo.initIsNFTFlag
 	issueSequenceInBlock uint32
 }
 
+type assetConstInfoRecord struct {
```
alexeykiselev, line 55: Why keep both types `assetConstInfo` and `assetConstInfoRecord`? May be remove one? Or convert the old one in CBOR record?

pkg/state/fee_validation.go
```diff
@@ -48,8 +48,8 @@ type feeValidationParams struct {
 }
 
 type assetParams struct {
-	quantity   int64
-	decimals   int32
+	quantity   uint64
```
alexeykiselev, line 51: I think that `int` types were chosen intentionally for better compatibility with Scala implementation.

pkg/state/fee_validation.go
```diff
@@ -95,16 +95,16 @@ func minFeeInUnits(params *feeValidationParams, tx proto.Transaction) (uint64, e
 	fee := baseFee
 	switch txType {
 	case proto.IssueTransaction:
-		var asset assetParams
+		var ap assetParams
 		switch itx := tx.(type) {
 		case *proto.IssueWithSig:
-			asset = assetParams{int64(itx.Quantity), int32(itx.Decimals), itx.Reissuable}
```
alexeykiselev, line 101: May be better to add checks that values fit in `int` types here?

alexeykiselev (changes requested):

alexeykiselev (approved):

nickeskov:
Testnet import on commit [d4a85d0](https://github.com/wavesplatform/gowaves/pull/1333/commits/d4a85d0e6264d064fc98735209aaa94939391523) is ok

alexeykiselev (approved):

PR #1337, opened 2024-02-14
Support block snapshot hash in 'statehash' utility

nickeskov:


cmd/statehash/statehash.go
```diff
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/pkg/errors"
```
alexeykiselev, line 14: We are trying to get rid of this unsupported package and planing to remove it completely form entire project and you introduce it here :)

alexeykiselev (changes requested):

cmd/statehash/statehash.go
nickeskov: Fixed

alexeykiselev (approved):

alexeykiselev (approved):

PR #1340, opened 2024-02-15
Generate fix snapshots and apply them at initial snapshot stage

nickeskov:
This PR **breaks compatibility** with the previous state version.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -1316,11 +1316,11 @@ func (s *stateManager) doBlockchainFixActionIfNeeded(
 		return nil, err
 	}
 	if cancelLeases {
-		clErr := s.cancelLeases(blockchainHeight, lastBlock) // TODO: should return atomic snapshots
+		fixes, clErr := s.cancelLeases(blockchainHeight, lastBlock)
 		if clErr != nil {
 			return nil, clErr
 		}
-		// TODO: fixSnapshots must be initialized here
+		fixSnapshots = fixes
 	}
 	resetStolenAliases, err := s.needToResetStolenAliases(blockchainHeight)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


nickeskov:
Testnet import on [b1ec5f3](https://github.com/wavesplatform/gowaves/pull/1340/commits/b1ec5f3767a7bf149f3877005485f9cb92f1e4a5) is ok

nickeskov:
Testnet import on [b5c47af](https://github.com/wavesplatform/gowaves/pull/1340/commits/b5c47af630c22f7d96e8da7d1aa63edc75c5591a) is ok

nickeskov:
Mainnet import on [b5c47af](https://github.com/wavesplatform/gowaves/pull/1340/commits/b5c47af630c22f7d96e8da7d1aa63edc75c5591a) is ok

pkg/state/state.go
```diff
@@ -1534,16 +1642,40 @@ func (s *stateManager) addBlocks() (*proto.Block, error) {
...
-		if addErr := s.addNewBlock(block, lastAppliedBlock, chans, blockchainCurHeight, sh); addErr != nil {
+		addErr := s.addNewBlock(block, lastAppliedBlock, chans, blockchainCurHeight, fixSnapshotsToInitialHash, sh)
+		if addErr != nil {
 			return nil, addErr
 		}
+		blockchainCurHeight++ // we've just added a new block, so we have a new height
 
-		if s.needToFinishVotingPeriod(blockchainCurHeight + 1) {
+		if s.needToFinishVotingPeriod(blockchainCurHeight) {
 			// If we need to finish voting period on the next block (h+1) then
```
alexeykiselev, line 1659: I don't get all those hight manipulations. The `blockchainCurHeight + 1` is the hight of current block. So we checking if the voting period ended on previous block (on blockchain height) we set the protobuf activation flag on for the current block and following blocks. Correct? 
But how it was deserialised without errors on line 1606 if this block is a protobuf block and the flag was not set?

alexeykiselev: And the comment is not correct then.

alexeykiselev (approved):

PR #1346, opened 2024-02-22
Transaction version validation restored after activation on feature 22.

alexeykiselev:


pkg/grpc/server/transactions_api.go
```diff
@@ -222,7 +223,12 @@ func (s *Server) Broadcast(ctx context.Context, tx *pb.SignedTransaction) (out *
 	if err != nil {
 		return nil, apiError(err)
 	}
-	t, err = t.Validate(s.scheme)
+	lightNodeActivated, err := s.state.IsActivated(int16(settings.LightNode))
```
nickeskov, line 226: `IsActivated` used `features.newestIsActivated`. Is it correct to use `Newest` methods here?

pkg/node/fsm/fsm_common.go
```diff
@@ -119,7 +120,12 @@ func tryBroadcastTransaction(
 			}
 		}()
 	}
-	if _, err = t.Validate(baseInfo.scheme); err != nil {
+	lightNodeActivated, err := baseInfo.storage.IsActivated(int16(settings.LightNode))
```
nickeskov, line 123: The same as above

pkg/state/appender.go
```diff
@@ -390,7 +390,11 @@ func (a *txAppender) verifyWavesTxSigAndData(tx proto.Transaction, params *appen
...
 		// because they have to be checked in each validateNextTx() anyway.
-		return checkTx(tx, checkTxSig, checkOrder1, checkOrder2, a.settings.AddressSchemeCharacter)
+		vp := proto.TransactionValidationParams{
+			Scheme:       a.settings.AddressSchemeCharacter,
+			CheckVersion: params.lightNodeActivated,
+		}
+		return checkTx(tx, checkTxSig, checkOrder1, checkOrder2, vp)
 	}
 	// Send transaction for validation of transaction's data correctness (using tx.Validate() method)
 	// and simple cryptographic signature verification (using tx.Verify() and PK).
```
nickeskov, line 400: I think the tasks for tx and block verifying should be separated into different structs.
Because the `checkVersion` field initialization is missed below.

pkg/proto/transactions.go
```diff
@@ -201,7 +208,7 @@ type Transaction interface {
 	// Validate checks that all transaction fields are valid.
 	// This includes ranges checks, and sanity checks specific for each transaction type:
 	// for example, negative amounts for transfers.
-	Validate(scheme Scheme) (Transaction, error)
+	Validate(params TransactionValidationParams) (Transaction, error)
```
nickeskov, line 211: On my mind `TransactionValidationParams` **may** be an interface for flexibility.

pkg/proto/transactions.go
```diff
@@ -532,7 +539,7 @@ func NewUnsignedGenesis(recipient WavesAddress, amount, timestamp uint64) *Genes
 }
 
 // Validate checks the validity of transaction parameters and it's signature.
-func (tx *Genesis) Validate(scheme Scheme) (Transaction, error) {
+func (tx *Genesis) Validate(params TransactionValidationParams) (Transaction, error) {
 	if tx.Version < 1 || tx.Version > MaxGenesisTransactionVersion {
```
nickeskov, line 543: I think max version of any transaction should be presented as a constant.

pkg/proto/transactions_with_proofs.go
```diff
@@ -187,8 +187,9 @@ func NewUnsignedIssueWithProofs(v byte, senderPK crypto.PublicKey, name, descrip
 	return &IssueWithProofs{Type: IssueTransaction, Version: v, Script: script, Issue: i}
 }
 
-func (tx *IssueWithProofs) Validate(_ Scheme) (Transaction, error) {
-	if tx.Version < 2 || tx.Version > MaxIssueTransactionVersion {
+func (tx *IssueWithProofs) Validate(params TransactionValidationParams) (Transaction, error) {
+	if tx.Version < 2 || params.CheckVersion && tx.Version > MaxIssueTransactionVersion ||
+		!params.CheckVersion && tx.Version > MaxUncheckedTransactionVersion {
 		return tx, errors.Errorf("unexpected version %d for IssueWithProofs", tx.Version)
 	}
```
nickeskov, line 194: These checks are similar. The common function can be used for tx version check. Take a look:
```go
func checkTxVersion(tx Transaction, expectedV byte, checkVersion bool) error {
	v := tx.GetVersion()
	if v < 1 || checkVersion && v > expectedV || !checkVersion && v > MaxUncheckedTransactionVersion {
		return errors.Errorf("unexpected version for (%T): want %d, got %d", tx, expectedV, v)
	}
	return nil
}
```

nickeskov (changes requested):

pkg/proto/transactions_with_proofs.go
alexeykiselev: It's code generation made by hand. No need to add function.

pkg/proto/transactions.go
alexeykiselev: Don't get you. There is a constant for max version of every transaction type.

pkg/grpc/server/transactions_api.go
alexeykiselev: I think it doesn't matter. It's a user API, transaction doesn't get to blockchain directly from here, it will be validated once more (at least) in UTX during block generation.

pkg/node/fsm/fsm_common.go
alexeykiselev: Here it's also OK, transaction gets into UTX from  here. Think of it as a pre validation. If a transaction gets invalidated because of activation of feature 22 what is the probability for it to became valid? I thinks it's zero.

pkg/state/appender.go
alexeykiselev: Yes, I see, fixed.

pkg/proto/transactions_with_proofs.go
nickeskov: What do you mean "by hand"? The code in this file is not generated. Hence, the whole code in this file can't be considered as generated. Generated code (imho) should be placed into a separate file.

pkg/proto/transactions.go
nickeskov: For testing, for example. For now it doesn't matter. It's only about potential

pkg/node/fsm/fsm_common.go
nickeskov: Okay

pkg/state/verifier.go
```diff
@@ -41,158 +42,150 @@ func (ch *verifierChans) closeAndWait() error {
...
+	if ok, err := vf(); !ok {
+		if err != nil {
+			return errs.Extend(err, fmt.Sprintf("%s transaction signature verification failed", name))
 		}
+		return errs.NewTxValidationError(fmt.Sprintf("%s transaction signature verification failed", name))
+	}
+	return nil
+}
+
+func verifyTransactionsWithProofs(tx proto.Transaction, scheme proto.Scheme) error {
```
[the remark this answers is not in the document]
alexeykiselev: Removed

pkg/proto/transactions_with_proofs.go
alexeykiselev: Don't know what to say, I mean that this code resembles generated. Of course, I didn't tell that it generated.

nickeskov (approved):

PR #1363, opened 2024-03-12
Generate keyblock with snapshot hash after LightNode feature

Anton-Rampage:


alexeykiselev (dismissed):

pkg/node/fsm/ng_state.go
```diff
@@ -296,6 +296,23 @@ func (a *NGState) mineMicro(
...
+	if err != nil {
+		return a, nil, a.Errorf(err)
+	}
+	if ok {
+		sh, errSh := a.baseInfo.storage.SnapshotStateHashAtHeight(lastBlockHeight)
+		if errSh != nil {
+			return a, nil, a.Errorf(errSh)
+		}
+		micro.StateHash = &sh
+	}
```
nickeskov, line 315: The result block will be invalid because it's signed and generated in `MicroMiner.Micro` method. `StateHash` should be calculated there and the block should be signed when this field is set.

Maybe modify `CreateBlock` constructor and add new fields to the its arguments?

pkg/node/fsm/wait_micro_snapshot.go
```diff
@@ -162,8 +162,21 @@ func (a *WaitMicroSnapshotState) checkAndAppendMicroBlock(
...
+	// here the blockchainHeight is equal to lastBlockHeight because we are appending a microblock to the last block
+	lastBlockHeight := blockchainHeight
+	ok, err = a.baseInfo.storage.IsActiveLightNodeNewBlocksFields(lastBlockHeight)
+	if err != nil {
+		return nil, a.Errorf(err)
+	}
+	if ok {
+		newBlock.StateHash = micro.StateHash
+	}
+
```
nickeskov, line 179: Now I think that all these logic is not necessary because the `StateHash` field existence should be checked in `ApplyMicroWithSnapshots`

pkg/node/fsm/ng_state.go
nickeskov, line 1: What about `checkAndAppendMicroBlock` method? I don't see that `StateHash` field of the passed micro block is even touched there.
I think that in `NG` mode after `Light Node` feature micro blocks can't be handled and applied because of that.
The 'StateHash' field of the result block must be set before `BlockID` generation and signing.

nickeskov (changes requested):

nickeskov (changes requested):

pkg/state/appender.go
```diff
@@ -853,42 +866,62 @@ func (a *txAppender) createCheckerInfo(params *appendBlockParams) (*checkerInfo,
...
 	// create the initial snapshot
 	initialSnapshot, err := a.txHandler.tp.createInitialBlockSnapshot(minerAndRewardDiff.balancesChanges())
 	if err != nil {
-		return crypto.Digest{}, errors.Wrap(err, "failed to create initial snapshot")
+		return txSnapshot{}, crypto.Digest{}, errors.Wrap(err, "failed to create initial snapshot")
 	}
 
-	currentBlockHeight := blockInfo.Height
-
+	// TODO: is it necesary to perform here any actions with diff storage??
```
nickeskov, line 904: Still TODO. What do you thing, @Anton-Rampage ?
Should it be removed or fixed (TODO looks like a question)?

nickeskov (commented):
LGTM, though take a look at the comments.

pkg/node/fsm/ng_state.go
```diff
@@ -244,6 +245,14 @@ func (a *NGState) MicroBlock(p peer.Peer, micro *proto.MicroBlock) (State, Async
...
 		a.baseInfo.MicroBlockCache.AddMicroBlock(block.BlockID(), micro)
 		a.blocksCache.AddBlockState(block)
+		a.baseInfo.scheduler.Reschedule()
+		// Notify all connected peers about new microblock, send them microblock inv network message
+		if inv, ok := a.baseInfo.MicroBlockInvCache.Get(block.BlockID()); ok {
+			//TODO: We have to exclude from recipients peers that already have this microblock
+			if err = broadcastMicroBlockInv(a.baseInfo, inv); err != nil {
+				return a, nil, a.Errorf(errors.Wrap(err, "failed to handle microblock message"))
+			}
+		}
```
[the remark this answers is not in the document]
nickeskov: Won't fix

PR #1411, opened 2024-05-19
Fix wallet's 'show' option.

nickeskov:
Now 'wallet' don't consider any seed as a valid 32 byte digest.
This PR fixes bug in #980.


.golangci-strict.yml
```diff
@@ -155,7 +155,7 @@ linters-settings:
   nolintlint:
     # Exclude following linters from requiring an explanation.
     # Default: []
-    allow-no-explanation: [ funlen, gocognit, lll ]
+    allow-no-explanation: [ funlen, gocognit, lll, forbidigo ]
```
alexeykiselev, line 158: I think it's not a very good move. By adding this exception we allow debug prints to stay in the code. I think, it's better to replace all `fmt.Print*` with `fmt.Fprintf(os.Stderr...` because it's the correct way to print user messages in CLI utilities. Or use `log`.

cmd/wallet/main.go
```diff
@@ -167,21 +174,21 @@ func showWallet(walletPath string, scheme proto.Scheme) error {
...
+		if genErr != nil {
+			return errors.Wrap(genErr, "failed to receive wallet's credentials")
 		}
 		fmt.Println()
-		fmt.Printf("Account number: %d\n", i)
-		fmt.Printf("Account seed:   %s\n", accountSeedDigest.String())
-		fmt.Printf("Public Key:     %s\n", pk.String())
-		fmt.Printf("Secret Key:     %s\n", sk.String())
-		fmt.Printf("Address:        %s\n", address.String())
+		fmt.Printf("Account number:         %d\n", i)                //nolint:forbidigo
```
alexeykiselev, line 183: Why `nolint` left in code?

alexeykiselev (changes requested):

alexeykiselev (approved):

PR #1497, opened 2024-09-16
Use 'securego/gosec@5f3194b' action

nickeskov:
Use action by commit sha '5f3194b' due to the securego/gosec#1219 issue.


nickeskov:
The problem still exists. Closing.

PR #1551, opened 2024-11-29
Peer address parsing refactoring

nickeskov:


pkg/proto/proto.go
```diff
@@ -737,26 +729,71 @@ func (a *IpPort) String() string {
...
+			continue
+		}
+		if ip.To4() == nil { // for now we support only IPv4
+			ips = append(ips[:i], ips[i+1:]...) // remove non-IPv4 address, assume that count of addresses is small
+		}
+	}
+	return ips
+}
+
+func resolveHostToIPv4(host string) (net.IP, error) {
```
alexeykiselev, line 752: Picking a random address isn’t what we need here. The idea of using hostnames for seed peers is that one hostname can point to a list of current seed peers. So, the resulting list should include all the IP addresses resolved from the seed peer’s hostname.

alexeykiselev (changes requested):

pkg/proto/proto_test.go
```diff
@@ -277,10 +279,67 @@ func TestPeerInfoMarshalJSON(t *testing.T) {
...
+		{"34.253.153.4:6868", PeerInfo{net.IPv4(34, 253, 153, 4), 6868}, ""},
+		{
+			"34.444.153.4:6868",
+			PeerInfo{},
+			"failed to resolve host: failed to resolve host \"34.444.153.4\": lookup 34.444.153.4: no such host",
+		},
+		{
+			"jfhasjdhfkmnn:6868",
+			PeerInfo{},
+			"failed to resolve host: failed to resolve host \"jfhasjdhfkmnn\": lookup jfhasjdhfkmnn: no such host",
```
alexeykiselev, line 296: This test fails we different error message, probably because of Docker:
```
--- FAIL: TestNewPeerInfoFromString (0.01s)
    --- FAIL: TestNewPeerInfoFromString/case_2 (0.01s)
        proto_test.go:336: 
            	Error Trace:	/home/runner/work/gowaves/gowaves/pkg/proto/proto_test.go:336
            	Error:      	Error message not equal:
            	            	expected: "failed to resolve host: failed to resolve host \"jfhasjdhfkmnn\": lookup jfhasjdhfkmnn: no such host"
            	            	actual  : "failed to resolve host: failed to resolve host \"jfhasjdhfkmnn\": lookup jfhasjdhfkmnn on 127.0.0.53:53: server misbehaving"
            	Test:       	TestNewPeerInfoFromString/case_2
```

alexeykiselev (changes requested):

pkg/proto/proto.go
nickeskov: Fixed by https://github.com/wavesplatform/gowaves/pull/1551/commits/f03d7020a4a7c4fb7867ac23aba872951f9422ee

pkg/proto/proto_test.go
nickeskov: Fixed by https://github.com/wavesplatform/gowaves/pull/1551/commits/b3a188c8888f81f06ff2e5473693f5aa3badd9f8

alexeykiselev (approved):

PR #1617, opened 2025-02-28
fix: fix incorrect nil return value

huochexizhan:
Because err has been checked before and return as != nil.

Err here must be nil, and `errCF` should actually be return.


nickeskov (approved):

huochexizhan:
Thanks for your review.

alexeykiselev (approved):

PR #1624, opened 2025-03-09
refactor: replace golang.org/x/exp with stdlib

StevenMia:
Since Go 1.21, the functions in x/exp used here can already be replaced by functions from the standard library.


nickeskov (approved):

nickeskov:
Thanks!

PR #1637, opened 2025-03-18
Fix semgerp issues created by public rulesets

nickeskov:


.github/workflows/deploy_node.yml
```diff
@@ -43,11 +43,14 @@ jobs:
   check-inputs:
     name: check-inputs
     runs-on: ubuntu-latest
+    environment:
```
nickeskov, line 46: Must be env

nickeskov: Fixed

alexeykiselev (approved):

PR #1640, opened 2025-03-21
fix: fix the wrong error return value

wellweek:
Fix the wrong error return value

Inspired by https://github.com/wavesplatform/gowaves/pull/1617


nickeskov (approved):

alexeykiselev (approved):

wellweek:
Thanks for your review.

PR #1645, opened 2025-03-28
refactor: use the built-in max/min to simplify the code

pkucode:
Use the built-in max/min from the standard library in Go 1.21 to simplify the code.  https://pkg.go.dev/builtin@go1.21.0#max


nickeskov:
Hello! Thanks for your contribution.
From my point of view, these changes cannot be considered a simplification. Please find more similar cases where the code can actually be simplified, and then we can consider merging your PR!
For now, I’m closing this.

PR #1672, opened 2025-04-28
Added l2-communication proto scheme

esuwu:


pkg/grpc/l2/blockchain_info/blockchain_info.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package blockchain_info;
+option java_package = "com.wavesplatform.protobuf";
+option csharp_namespace = "Waves";
```
alexeykiselev, line 5: I think options for Java and C# can be removed if it's possible.

alexeykiselev (approved):

PR #1709, opened 2025-06-02
Updated metrics

esuwu:


pkg/node/fsm/ng_state.go
```diff
@@ -203,7 +207,11 @@ func (a *NGState) Block(peer peer.Peer, block *proto.Block) (State, Async, error
 func (a *NGState) MinedBlock(
 	block *proto.Block, limits proto.MiningLimits, keyPair proto.KeyPair, vrf []byte,
 ) (State, Async, error) {
-	metrics.FSMKeyBlockGenerated("ng", block)
+	height, heightErr := a.baseInfo.storage.Height()
```
alexeykiselev, line 219: Shouldn't we add 1 to height here? Please, check how it made in Scala.

pkg/node/fsm/sync_state.go
```diff
@@ -196,8 +201,13 @@ func (a *SyncState) BlockSnapshot(
 func (a *SyncState) MinedBlock(
 	block *proto.Block, limits proto.MiningLimits, keyPair proto.KeyPair, vrf []byte,
 ) (State, Async, error) {
-	metrics.FSMKeyBlockGenerated("sync", block)
+	height, heightErr := a.baseInfo.storage.Height()
```
alexeykiselev, line 199: Same question with height here. But for now, generation in the Sync state has low priority.

pkg/node/fsm/wait_snapshot_state.go
```diff
@@ -108,6 +108,10 @@ func (a *WaitSnapshotState) BlockSnapshot(
 	}
 
 	defer a.cleanupBeforeTransition()
+	height, heightErr := a.baseInfo.storage.Height()
```
alexeykiselev, line 111: Same concern about height here.

pkg/node/fsm/wait_snapshot_state.go
```diff
@@ -117,8 +121,8 @@ func (a *WaitSnapshotState) BlockSnapshot(
 		zap.S().Errorf("%v", a.Errorf(errors.Wrapf(err, "Failed to apply block %s", a.blockWaitingForSnapshot.BlockID())))
 		return processScoreAfterApplyingOrReturnToNG(a, a.baseInfo, a.receivedScores, a.blocksCache)
 	}
-
-	metrics.FSMKeyBlockApplied("ng", a.blockWaitingForSnapshot)
+	metrics.BlockApplied(a.blockWaitingForSnapshot, height)
```
alexeykiselev: We report block application here, but snapshot can be applied after a micro block also, as I understand. What we should report here?

esuwu: Changed it

pkg/node/fsm/wait_snapshot_state.go
esuwu: It seems to me that this is the correct way to report height after discussion, but i'll double check it anyway

pkg/metrics/metrics.go
```diff
@@ -167,89 +187,36 @@ func BlockMined(block *proto.Block, height proto.Height) {
...
-	if rep == nil {
-		return
-	}
-	t := emptyTags().node().fsm(fsm).microblock().generated()
-	f := emptyFields().blockID(microblock.TotalBlockID).referenceID(microblock.Reference).signature(microblock.TotalResBlockSigField)
-	reportFSM(t, f)
-}
-
-func FSMMicroBlockDeclined(fsm string, microblock *proto.MicroBlock, err error) {
+func MicroBlockMined(mb *proto.MicroBlock) {
```
alexeykiselev: Here we should report total transactions count of the block, not micro-block. Looks like we have to add separate parameter here.

esuwu: Fixed

pkg/node/fsm/sync_state.go
esuwu: Fixed

pkg/node/fsm/ng_state.go
esuwu: Fixed

alexeykiselev (dismissed):

alexeykiselev (approved):

PR #1723, opened 2025-06-16
Metrics report fsm channel length

esuwu:


pkg/node/node.go
```diff
@@ -151,12 +152,21 @@ func (a *Node) logErrors(err error) {
 	}
 }
 
+func monitorChanLength(messageCh <-chan peer.ProtoMessage) {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
```
nickeskov: Who stops the ticker? Looks like that this function has infinite loop.

nickeskov: Also, `metrics.FSMChannelLength(len(messageCh))` can be placed inside `Node.runInternalMetrics` loop.

PR #1741, opened 2025-06-30
Auto fixes from zizmor applied.

alexeykiselev:


.github/workflows/run_itests.yml
```diff
@@ -21,8 +23,11 @@ jobs:
     steps:
       - name: Check inputs
         run: |
-          echo "${{ env.ITEST_TYPE }}" | grep -E '^(itest|itest-race|smoke)$' || exit 1 # check if the itest type is valid
-          echo "${{ env.ENVIRONMENT }}" | grep -E '^(^$|ITests)$' || exit 1 # check if the environment is valid
+          echo "${ITEST_TYPE}" | grep -E '^(itest|itest-race|smoke)$' || exit 1 # check if the itest type is valid
+          echo "${ENVIRONMENT}" | grep -E '^(^$|ITests)$' || exit 1 # check if the environment is valid
+        env:
+          ITEST_TYPE: ${{ env.ITEST_TYPE }}
+          ENVIRONMENT: ${{ env.ENVIRONMENT }}
```
nickeskov, line 30: Maybe move `env` section before `run` (just for clarity)?

.github/workflows/run_itests.yml
```diff
@@ -50,7 +57,9 @@ jobs:
         run: go mod vendor
 
       - name: Tests
-        run: make ${{ env.MAKEFILE_TARGET }} # run the integration tests
+        run: make ${MAKEFILE_TARGET} # run the integration tests
+        env:
+          MAKEFILE_TARGET: ${{ env.MAKEFILE_TARGET }}
```
nickeskov, line 62: And here

.github/workflows/itests.yml
```diff
@@ -52,12 +54,16 @@ jobs:
           fi
           echo "--> CAN_RUN_ITEST=$CAN_RUN_ITEST" # for debugging
           echo "CAN_RUN_ITEST=$CAN_RUN_ITEST" >> "$GITHUB_OUTPUT"
-
+        env:
```
nickeskov, line 57: And also here

.github/workflows/run_itests.yml
alexeykiselev: I think it's irrelevant, they are not executed in any particular order. As I understand it, env is just property of an action and has to be resolved before execution of the action.

.github/workflows/run_itests.yml
alexeykiselev: By the way, in the remediation section https://docs.zizmor.sh/audits/#remediation_18 they propose to put `env` after the `run`.

PR #1817, opened 2025-09-04
UTX clean-up reimplemented with in-place transactions validation.

alexeykiselev:
Pool item type heapItem added to store index of item in the heap. Item index management implemented in heap.Interface implementation. AllTransactions function now return slice of proto.Transaction. Function Clean added to UtxPool interface and implemented. Interface function Count renamed to Len, duplicated method removed from implementation.

### DIFF
--- a/pkg/miner/utxpool/pool.go
+++ b/pkg/miner/utxpool/pool.go
@@ -2,33 +2,47 @@ package utxpool
 
 import (
 	"container/heap"
+	"context"
 	"fmt"
+	"log/slog"
 	"sync"
 
 	"github.com/mr-tron/base58"
 	"github.com/pkg/errors"
 
 	"github.com/wavesplatform/gowaves/pkg/crypto"
+	"github.com/wavesplatform/gowaves/pkg/logging"
 	"github.com/wavesplatform/gowaves/pkg/proto"
 	"github.com/wavesplatform/gowaves/pkg/settings"
 	"github.com/wavesplatform/gowaves/pkg/types"
 )
 
-type transactionsHeap []*types.TransactionWithBytes
+type heapItem struct {
+	tx    *types.TransactionWithBytes
+	index int // The index of the item in the heap.
+}
+
+type transactionsHeap []*heapItem
 
 func (a transactionsHeap) Len() int { return len(a) }
 
 func (a transactionsHeap) Less(i, j int) bool {
 	// skip division by zero, check it when we add transaction
-	return a[i].T.GetFee()/uint64(len(a[i].B)) > a[j].T.GetFee()/uint64(len(a[j].B))
+	return a[i].tx.T.GetFee()/uint64(len(a[i].tx.B)) > a[j].tx.T.GetFee()/uint64(len(a[j].tx.B))
 }
 
 func (a transactionsHeap) Swap(i, j int) {
 	a[i], a[j] = a[j], a[i]
+	a[i].index = i
+	a[j].index = j
 }
 
 func (a *transactionsHeap) Push(x any) {
-	item := x.(*types.TransactionWithBytes)
+	item, ok := x.(*heapItem)
+	if !ok {
+		return // Don't put anything invalid in the heap.
+	}
+	item.index = len(*a)
 	*a = append(*a, item)
 }
 
@@ -36,6 +50,7 @@ func (a *transactionsHeap) Pop() any {
 	old := *a
 	n := len(old)
 	item := old[n-1]
+	item.index = -1 // For safety, mark as no longer in the heap.
 	*a = old[0 : n-1]
 	return item
 }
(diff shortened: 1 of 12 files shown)


pkg/miner/utxpool/pool.go
```diff
@@ -2,40 +2,55 @@ package utxpool
...
 	a[i], a[j] = a[j], a[i]
+	a[i].index = i
+	a[j].index = j
 }
 
 func (a *transactionsHeap) Push(x any) {
-	item := x.(*types.TransactionWithBytes)
+	item, ok := x.(*heapItem)
+	if !ok {
+		return // Don't put anything invalid in the heap.
```
nickeskov: Better to panic in this case

pkg/miner/utxpool/pool.go
```diff
@@ -59,58 +74,96 @@ func New(sizeLimit uint64, validator Validator, settings *settings.BlockchainSet
...
+	}
 	return res
 }
 
+func (a *UtxImpl) Clean(ctx context.Context, shouldDrop func(tx proto.Transaction) bool) (int, int) {
+	// Take a snapshot of the current transactions to avoid holding the lock for too long.
+	a.mu.Lock()
+	snapshot := make([]*heapItem, a.transactions.Len())
+	copy(snapshot, a.transactions)
+	a.mu.Unlock()
```
nickeskov, line 99: Maybe create internal `map` method with automatic lock and unlock?
For example:
```go
func (a *UtxImpl) threadSafe(f func()) {
	a.mu.Lock()
	defer a.mu.Unlock()
	f()
}
```

pkg/miner/utxpool/pool.go
```diff
@@ -2,40 +2,55 @@ package utxpool
...
+	}
+	item.index = len(*a)
 	*a = append(*a, item)
 }
 
 func (a *transactionsHeap) Pop() any {
 	old := *a
 	n := len(old)
 	item := old[n-1]
+	item.index = -1 // For safety, mark as no longer in the heap.
```
nickeskov, line 59: I think this also can be helpful for GC:
```go
	old[n-1] = nil  // avoid memory leak
```

pkg/miner/utxpool/pool.go
```diff
@@ -59,58 +74,96 @@ func New(sizeLimit uint64, validator Validator, settings *settings.BlockchainSet
...
+	// Take a snapshot of the current transactions to avoid holding the lock for too long.
+	a.mu.Lock()
+	snapshot := make([]*heapItem, a.transactions.Len())
+	copy(snapshot, a.transactions)
+	a.mu.Unlock()
+
+	// Check which transactions should be dropped.
+	var drop []*heapItem
+	checked := 0
+	q := a
```
nickeskov: Is it necessary?

pkg/miner/utxpool/pool.go
```diff
@@ -59,58 +74,96 @@ func New(sizeLimit uint64, validator Validator, settings *settings.BlockchainSet
...
+	}
+
+	// Now remove the dropped transactions from the heap.
+	a.mu.Lock()
+	for _, it := range drop {
+		if it.index >= 0 {
+			heap.Remove(&a.transactions, it.index)
+		}
+	}
+	q.mu.Unlock()
```
nickeskov: `a.mu.Lock()` and `q.mu.Unlock()` looks a bit unsafe.

pkg/miner/utxpool/pool.go
```diff
@@ -59,58 +74,96 @@ func New(sizeLimit uint64, validator Validator, settings *settings.BlockchainSet
...
+		res[i] = it.tx.T
+	}
 	return res
 }
 
+func (a *UtxImpl) Clean(ctx context.Context, shouldDrop func(tx proto.Transaction) bool) (int, int) {
+	// Take a snapshot of the current transactions to avoid holding the lock for too long.
+	a.mu.Lock()
+	snapshot := make([]*heapItem, a.transactions.Len())
+	copy(snapshot, a.transactions)
```
nickeskov, line 98: `slices.Clone`

pkg/miner/utxpool/pool.go
```diff
@@ -59,58 +74,96 @@ func New(sizeLimit uint64, validator Validator, settings *settings.BlockchainSet
...
 	a.mu.Lock()
 	defer a.mu.Unlock()
-	res := make([]*types.TransactionWithBytes, len(a.transactions))
-	copy(res, a.transactions)
+	items := make([]*heapItem, a.transactions.Len())
+	copy(items, a.transactions)
+	res := make([]proto.Transaction, len(items))
+	for i, it := range items {
+		res[i] = it.tx.T
+	}
```
nickeskov: It's not necessary to copy all transactions pointers into a temporary slice. Btw, maybe change regular mutex to a read-write mutex? What do you think?
```suggestion
	res := make([]proto.Transaction, a.transactions.Len())
	for i, it := range a.transactions {
		res[i] = it.tx.T
	}
	return res
```

pkg/miner/utxpool/pool.go
```diff
@@ -121,53 +174,58 @@ func (a *UtxImpl) addWithBytesOptValidation(
...
 		}
 	}
-	tb := &types.TransactionWithBytes{
-		T: t,
-		B: b,
+	it := &heapItem{
+		tx: &types.TransactionWithBytes{
+			T: tx,
+			B: b,
+		},
```
nickeskov, line 200: Maybe also set index to `-1` as initial state?

pkg/miner/utxpool/pool.go
```diff
@@ -184,16 +242,28 @@ func (a *UtxImpl) ExistsByID(id []byte) bool {
...
-		return tb
+
+	if a.transactions.Len() == 0 {
+		return nil
 	}
-	return nil
+	it, ok := heap.Pop(&a.transactions).(*heapItem)
+	if !ok {
+		return nil
+	}
```
nickeskov: Also can be a reason to panic because if it happens than it's 100% bug.

pkg/state/stateerr/errors.go
```diff
@@ -54,8 +54,6 @@ func (err StateError) Unwrap() error {
 func IsTxCommitmentError(err error) bool {
 	var stateErr StateError
 	switch {
-	case err == nil:
```
nickeskov, line 57: Can be simplified to simple `if` statement. Also `err == nil` can be removed in `IsNotFound`.

pkg/miner/utxpool/pool.go
alexeykiselev: Fixed

pkg/miner/utxpool/pool.go
alexeykiselev: Fixed

pkg/miner/utxpool/pool.go
alexeykiselev: No, it’s just an artifact by Copilot left behind. It became too verbose and tries to generate or modify entire functions. 
Removed.

pkg/miner/utxpool/pool.go
alexeykiselev: The same. Removed.

pkg/miner/utxpool/pool.go
alexeykiselev: Nice suggestion but `Clone` uses `append` to copy elements and it's documentation states that:

> The result may have additional unused capacity.

Which is bad because we are not going to append any items to the snapshot.
Left unchanged.

pkg/miner/utxpool/pool.go
alexeykiselev: Yes, good point, also added index check to `Push` function.

pkg/miner/utxpool/pool.go
alexeykiselev: Yes, good point, did like this.

pkg/miner/utxpool/pool.go
alexeykiselev: Yes, replaced with panic.

pkg/state/stateerr/errors.go
alexeykiselev: Fixed, and all file improved.

pkg/miner/utxpool/pool.go
alexeykiselev: Feels a bit like over-engineering to me.

PR #1834, opened 2025-09-22
New feature 25 "Deterministic Finality and Ride V9" added.

alexeykiselev:


nickeskov:
New `TODO`s brought by this changes
```sh
git log master..determenistic-finality-feature -G "TODO" -p | grep -E "\+.*TODO" | uniq
```
```
+		// TODO: remove height parameter from the following function. Micro-block mining operates only on current
+// TODO: Move JSON tags to GeneratorInfo structure.
+	// TODO: This would work only for height still present in history, once the height fell out of history, no
+		// TODO: remove height parameter from the following function. Micro-block mining is operates only current
+		//TODO: Implement. The implementation is possible for extended API only, when the generators are stored for
+		Ignore(BlockEndorsementEvent). // TODO: Consider adding endorsements to pool.
+			// TODO: maybe return an error, not log it?
+// TODO: what block ID should be provided: applying one or its parent?
+	// TODO check if generator is in the generator set.
+		// TODO: we should distinguish between block is full because of size or because of complexity
+	a.baseInfo.actions.SendEndorseBlock(blockEndorsement) // TODO should we send it out if conflicting?
+	// TODO here we somehow have to check that if endorsedBlock was N, we are at N+2.
+			parent.BlockID()) // TODO is it parent BlockID?
+		// TODO should we return for these errors too?
+	// TODO: compare with scala node results
+	}{ // TODO: compare with scala node results
+		// TODO handle Ethereum invoke expression tx
+		// TODO: Implement the case after this tx type supported.
+	// TODO: implement conversion after adding CommitToGenerationTransaction type to Ride V9.
+// TODO: Should be renamed to isNotFoundInHistoryOrEmptyHistory.
+	// TODO: But maybe it's better to calculate it from the block height?
+		//TODO: Add Deposit to the profile.
```

PR #1838, opened 2025-09-23
Bls aggregated sig refactoring

alexeykiselev:


pkg/crypto/bls/bls.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func Verify(pk PublicKey, msg []byte, sig Signature) (bool, error) {
+	cpk, err := pk.ToCIRCLPublicKey()
+	if err != nil {
+		return false, fmt.Errorf("failed to verify signature: %w", err)
+	}
+	if !cpk.Validate() {
+		return false, fmt.Errorf("failed to verify signature: invalid public key")
+	}
```
nickeskov: Done intentionally? For example, `VerifyAggregate` does not contain this lines. As I can remember we've discussed that this is necessary only for _proof-of-possession_ operation. Also some sources say that this operation can be slow.

pkg/crypto/bls/bls.go
```diff
@@ -0,0 +1,286 @@
...
+	copy(s[:], b[:PublicKeySize])
+	return nil
+}
+
+func (s Signature) String() string {
+	return base58.Encode(s[:])
+}
+
+func (s Signature) ShortString() string {
+	const ellipsis = 0x2026
```
nickeskov: Please, add a comment for the constant.

pkg/crypto/bls/bls.go
```diff
@@ -0,0 +1,286 @@
...
+
+func (k *SecretKey) Bytes() []byte {
+	return k[:]
+}
+
+func (k *SecretKey) String() string {
+	return base58.Encode(k[:])
+}
+
+func (k *SecretKey) ToCIRCLSecretKey() (*cbls.PrivateKey[cbls.G1], error) {
```
nickeskov: Guess this method should be package private. If if necessary only for tests, then do this operation in the `bls_test` package.

pkg/crypto/bls/bls.go
```diff
@@ -0,0 +1,286 @@
...
+	if l := len(b); l != SecretKeySize {
+		return sk, crypto.NewIncorrectLengthError("BLS SecretKey", l, SecretKeySize)
+	}
+	copy(sk[:], b[:SecretKeySize])
+	return sk, nil
+}
+
+// NewSecretKeyFromWavesSecretKey generates BLS secret key from Waves secret key.
+func NewSecretKeyFromWavesSecretKey(wavesSK crypto.SecretKey) (SecretKey, error) {
+	k, err := cbls.KeyGen[cbls.G1](wavesSK.Bytes(), nil, nil)
```
nickeskov: No salt?

alexeykiselev: Haven't investigated this, it was done like this before.

pkg/crypto/bls/bls.go
alexeykiselev: Added.

pkg/crypto/bls/bls.go
alexeykiselev: I removed it, actually the validation of the key always called inside CIRCL functions like Verify and Sign.
The only place we need to validate public key manually is in the PoPVerify and only to produce a meaningful error.

pkg/crypto/bls/bls.go
alexeykiselev: Made it private, and it's not used only in tests.

alexeykiselev: Rewrote the test, it was weird.

pkg/crypto/bls/bls.go
```diff
@@ -0,0 +1,283 @@
...
+
+// Sign calculates 96-byte compressed BLS signature over msg.
+// Default separation tag "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_" is used.
+func Sign(sk SecretKey, msg []byte) (Signature, error) {
+	csk, err := sk.toCIRCLSecretKey()
+	if err != nil {
+		return Signature{}, fmt.Errorf("failed to sign: %w", err)
+	}
+	s := cbls.Sign[cbls.G1](csk, msg)
+	return MustSignatureFromBytes(s), nil
```
nickeskov: Can be replaced to regular `NewSignatureFromBytes`

alexeykiselev: Yes, fixed.

PR #1852, opened 2025-10-06
Add network messages

esuwu:


pkg/node/fsm/ng_state.go
```diff
@@ -219,8 +275,28 @@
...
+	if !ok {
+		return a, nil, a.Errorf(errors.Errorf("verification of block %d failed", height))
+	}
+	canFinalize, err := a.baseInfo.storage.CalculateVotingFinalization(a.baseInfo.endorsements.GetEndorsers(), height, a.baseInfo.endorsements.GetGenerators())
+	if err != nil {
+		return a, nil, err
+	}
+	if err != nil {
+		return a, nil, a.Errorf(errors.Errorf("failed to calculate finalization voting %v", err))
+	}
```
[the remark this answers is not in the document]
nickeskov: @esuwu it's still valid.
Moreover, you're also using %v instead of %w.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+type key struct {
+	blockID       [32]byte
+	endorserIndex int32
+}
+
+func makeKey(blockID []byte, idx int32) key {
+	var k key
+	copy(k.blockID[:], blockID)
+	k.endorserIndex = idx
+	return k
```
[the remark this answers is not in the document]
nickeskov: Agree, maybe use `BlockID` type?

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+	it, ok := x.(*heapItemEndorsement)
+	if !ok {
+		panic(fmt.Sprintf("endorsementHeap.Push: unexpected type %T", x))
+	}
+	it.index = len(*h)
+	*h = append(*h, it)
+}
+
+func (h *endorsementHeap) Pop() any {
+	old := *h
```
nickeskov, line 57: What if `endorsementHeap` is `nil`?

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+
+type EndorseVerifier interface {
+	Verify(eb *proto.EndorseBlock) bool
+}
+
+type EndorsementPool struct {
+	mu                      sync.Mutex
+	seq                     uint64
+	countLimit              int
+	byKey                   map[key]*heapItemEndorsement
```
nickeskov: I think pointer is not necessary here, you can use `heapItemEndorsement` type as is.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+	h                       endorsementHeap
+	endorsersPublicKeyCache GeneratorsPublicKeysCache
+}
+
+func NewEndorsementPool() *EndorsementPool {
+	return &EndorsementPool{
+		countLimit: endorsementPoolLimit,
+		byKey:      make(map[key]*heapItemEndorsement),
+	}
+}
```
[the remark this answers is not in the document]
nickeskov: Yes, it should be initialized.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+	if e == nil || len(e.EndorsedBlockID) == 0 {
+		return errors.New("invalid endorsed block id")
+	}
+	k := makeKey(e.EndorsedBlockID, e.EndorserIndex)
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if _, exists := p.byKey[k]; exists {
+		return errors.New("duplicate endorsement")
```
nickeskov: Please, add endorser index and block ID to the error message.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+type endorsementHeap []*heapItemEndorsement
+
+func (h endorsementHeap) Len() int { return len(h) }
+
+// Less Oldest = smallest seq - floats to top.
+func (h endorsementHeap) Less(i, j int) bool {
+	return h[i].seq < h[j].seq
+}
```
nickeskov: Is it intended that endorsements are sorted by `seq`, not an `index`?

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+	return out
+}
+
+func (p *EndorsementPool) Finalize() proto.FinalizationVoting {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	return proto.FinalizationVoting{}
```
nickeskov: TODO?

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+	return nil
+}
+
+func (p *EndorsementPool) GetAll() []*proto.EndorseBlock {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	out := make([]*proto.EndorseBlock, 0, len(p.h))
+	for _, it := range p.h {
+		if it != nil {
```
nickeskov: Can `it` be `nil`?

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+		err := sig.UnmarshalJSON(heapItem.eb.Signature)
+		if err != nil {
+			return false, err
+		}
+		sigs = append(sigs, sig)
+		pks = append(pks, heapItem.endorserPK)
+	}
+	msg, err := p.h[0].eb.EndorsementMessage() // All messages are assumed the same.
+	if err != nil {
+		return false, err
```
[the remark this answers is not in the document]
nickeskov: Agree

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,229 @@
...
+	k := makeKey(it.eb.EndorsedBlockID, it.eb.EndorserIndex)
+	delete(p.byKey, k)
+	return it.eb
+}
+
+func (p *EndorsementPool) Verify() (bool, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	var sigs []bls.Signature
+	var pks []bls.PublicKey
```
nickeskov: You can pre-allocate memory for `sigs` and `pks`. I think optimistic approach here would be optimal.

pkg/node/fsm/fsm.go
```diff
@@ -396,6 +400,12 @@ func (f *FSM) BlockSnapshot(p peer.Peer, blockID proto.BlockID, snapshots proto.
 	return *asyncRes, err
 }
 
+func (f *FSM) BlockEndorsement(p peer.Peer, blockID proto.BlockID, snapshots proto.BlockSnapshot) (Async, error) {
+	asyncRes := &Async{}
+	err := f.fsm.Fire(BlockEndorsementEvent, asyncRes, p, blockID, snapshots)
```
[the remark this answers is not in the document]
nickeskov: Agree, looks very suspicious. Probably it will panic.

pkg/node/fsm/ng_state.go
```diff
@@ -205,6 +205,36 @@ func (a *NGState) Block(peer peer.Peer, block *proto.Block) (State, Async, error
...
+			endorsedMicroBlock.TotalBlockID.String(), endorsedMicroBlock.Reference.String(), top.BlockID().String())
+		return a, nil, proto.NewInfoMsg(err)
+	}
+
+	// TODO verify individual endorsements (signatures) using the endorser's public key.
+	// TODO validate whether the endorsement is not a duplicate
+
+	err = a.baseInfo.endorsements.Add(blockEndorsement)
+	if err != nil {
+		return nil, nil, errors.Errorf("failed to add an endorsement, %v", err)
```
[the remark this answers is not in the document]
nickeskov: That will panic 100%.
Also, use `<your msg>: %w` for wrapping errors.

pkg/node/fsm/ng_state.go
```diff
@@ -532,6 +586,21 @@ func initNGStateInFSM(state *StateData, fsm *stateless.StateMachine, info BaseIn
...
+			createPermitDynamicCallback(BlockEndorsementEvent, state, func(args ...any) (State, Async, error) {
+				a, ok := state.State.(*NGState)
+				if !ok {
+					return a, nil, a.Errorf(errors.Errorf(
+						"unexpected type '%T' expected '*NGState'", state.State))
+				}
+				endorse, ok := args[1].(*proto.EndorseBlock)
+				if !ok {
+					return a, nil, a.Errorf(errors.Errorf("unexpected type %T, expected *proto.EndorseBlock", args[1]))
+				}
```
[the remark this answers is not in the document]
nickeskov: Yep

pkg/node/fsm/fsm.go
```diff
@@ -396,6 +400,12 @@ func (f *FSM) BlockSnapshot(p peer.Peer, blockID proto.BlockID, snapshots proto.
 	return *asyncRes, err
 }
 
+func (f *FSM) BlockEndorsement(p peer.Peer, blockID proto.BlockID, snapshots proto.BlockSnapshot) (Async, error) {
```
nickeskov: Can find an `action` for this method for now.

pkg/node/fsm/ng_state.go
```diff
@@ -205,6 +205,36 @@ func (a *NGState) Block(peer peer.Peer, block *proto.Block) (State, Async, error
...
 
+func (a *NGState) BlockEndorsement(_ peer.Peer, blockEndorsement *proto.EndorseBlock) (State, Async, error) {
+	endorsedBlockID, err := proto.NewBlockIDFromBytes(blockEndorsement.EndorsedBlockID)
+	if err != nil {
+		return a, nil, a.Errorf(err)
+	}
+
+	endorsedMicroBlock, found := a.baseInfo.MicroBlockCache.GetBlock(endorsedBlockID)
+	if !found || endorsedMicroBlock == nil {
+		return a, nil, a.Errorf(errors.New("failed to find the microblock that was endorsed"))
```
nickeskov: Please, add `endorsedBlockID` to the error message.

pkg/node/fsm/ng_state.go
```diff
@@ -219,8 +249,26 @@ func (a *NGState) MinedBlock(
 	if heightErr != nil {
 		return a, nil, a.Errorf(heightErr)
 	}
+
+	ok, err := a.baseInfo.endorsements.Verify()
```
nickeskov: I think it's necessary to call only after corresponding feature activation.

pkg/node/fsm/ng_state.go
```diff
@@ -219,8 +249,26 @@ func (a *NGState) MinedBlock(
...
+	if err != nil {
+		return a, nil, a.Errorf(err)
+	}
+	if !ok {
+		return a, nil, a.Errorf(errors.Errorf("verification of block %d failed", height))
+	}
+	canFinalize, err := a.baseInfo.storage.CalculateVotingFinalization(a.baseInfo.endorsements.GetEndorsers(), height,
+		a.baseInfo.endorsements.GetGenerators())
+	if err != nil {
+		return a, nil, a.Errorf(errors.Errorf("failed to calculate finalization voting %v", err))
```
nickeskov: `<msg>: %w`

pkg/node/fsm/ng_state.go
```diff
@@ -239,6 +287,7 @@ func (a *NGState) MinedBlock(
 	slog.Info("Generated key block successfully applied to state", "state", a.String(),
 		"blockID", block.ID.String())
 
+	a.baseInfo.endorsements.CleanAll()
```
nickeskov, line 556: Not cache, all?

pkg/node/fsm/ng_state.go
```diff
@@ -303,6 +353,10 @@ func (a *NGState) mineMicro(
 		return a, nil, a.Errorf(errors.Wrap(err, "failed to generate microblock"))
 	}
 	metrics.MicroBlockMined(micro, block.TransactionCount)
+
+	endorsements := a.baseInfo.endorsements.GetAll()
+	micro.Endorsements = endorsements
```
nickeskov: Guess that should be places in `MicroMiner`.

pkg/grpc/protobuf-schemas
nickeskov, line 1: Update from specific branch?

pkg/state/state.go
```diff
@@ -3189,3 +3189,30 @@ func (s *stateManager) Close() error {
...
+	for _, endorser := range endorsers {
+		endorserRecipient := proto.NewRecipientFromAddress(endorser)
+		balance, err := s.GeneratingBalance(endorserRecipient, height)
+		if err != nil {
+			return false, err
+		}
+		endorsersGeneratingBalance += balance
+	}
+	// If endorsersBalance >= 2/3 totalGeneratingBalance
+	if 3*endorsersGeneratingBalance >= 2*totalGeneratingBalance {
```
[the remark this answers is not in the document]
nickeskov: Actually, this won't help. Better to use functions from `pkg/util/common/util.go`.

pkg/state/state.go
```diff
@@ -3189,3 +3189,30 @@ func (s *stateManager) Close() error {
 	}
 	return nil
 }
+
+func (s *stateManager) CalculateVotingFinalization(endorsers []proto.WavesAddress, height proto.Height,
```
nickeskov: Maybe rename to check-smth?
Also, I don't think that placing this logic in `stateManager` is a good idea. Probably it should be in another place, maybe in `consensus` package.
What do you think?

pkg/state/api.go
```diff
@@ -154,6 +154,11 @@ type StateInfo interface {
 
 	// SnapshotsAtHeight returns block snapshots at the given height.
 	SnapshotsAtHeight(height proto.Height) (proto.BlockSnapshot, error)
+
+	// CalculateVotingFinalization calculates whether the generating balance of endorsers at the previous block
+	// Exceeds the total generating balance of all committed generators for that block.
```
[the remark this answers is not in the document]
nickeskov: Which `height`: block or blockchain?

pkg/proto/microblock.go
```diff
@@ -32,6 +32,7 @@ type MicroBlock struct {
 	SenderPK              crypto.PublicKey
 	Signature             crypto.Signature
 	StateHash             *crypto.Digest // is nil before protocol version 1.5
+	Endorsements          []*EndorseBlock
```
nickeskov: Maybe also use `EndorseBlock` instead of `*EndorseBlock`?

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+
+type EndorseBlock struct {
+	EndorserIndex        int32  `json:"endorserIndex"`
+	FinalizedBlockID     []byte `json:"finalizedBlockID"`
+	FinalizedBlockHeight uint32 `json:"finalizedBlockHeight"`
+	EndorsedBlockID      []byte `json:"endorsedBlockId"`
+	Signature            []byte `json:"signature"`
+}
+
+func (e *EndorseBlock) Marshal() ([]byte, error) {
```
nickeskov, line 21: This method name is too wide. Maybe rename it to something specific? And, is it necessary?

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+	endBlockProto := e.ToProtobuf()
+	return endBlockProto.MarshalVTStrict()
+}
+
+// EndorsementMessage serializes endorsement structure into base58 message.
+func (e *EndorseBlock) EndorsementMessage() ([]byte, error) {
+	if len(e.FinalizedBlockID) == 0 || len(e.EndorsedBlockID) == 0 {
+		return nil, errors.New("invalid endorsement: missing block IDs")
+	}
```
nickeskov: Maybe return two different errors in this cases?

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+	// finalizedBlockId + 4 bytes height + endorsedBlockId
+	size := len(e.FinalizedBlockID) + heightSize + len(e.EndorsedBlockID)
+	buf := make([]byte, size)
+	// finalizedBlockId
+	copy(buf[0:len(e.FinalizedBlockID)], e.FinalizedBlockID)
+	// finalizedBlockHeight
+	binary.BigEndian.PutUint32(buf[len(e.FinalizedBlockID):len(e.FinalizedBlockID)+4], e.FinalizedBlockHeight)
+	// endorsedBlockId
+	copy(buf[len(e.FinalizedBlockID)+4:], e.EndorsedBlockID)
```
nickeskov: Magic number `4`, use `heightSize` constant.

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return []byte(base58.Encode(buf)), nil
```
nickeskov: That's not valid. `buf` must be returned as `EndorsementMessage` without encoding it to `base58`. Check scala node implementation ([reference](https://github.com/wavesplatform/Waves/blob/d3687e894de36fb02636e611c2e828ad17453b15/node/src/main/scala/com/wavesplatform/block/BlockEndorsement.scala#L19)).

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/mr-tron/base58"
+	"github.com/pkg/errors"
+
+	g "github.com/wavesplatform/gowaves/pkg/grpc/generated/waves"
+)
+
+const heightSize = 4
+
+type EndorseBlock struct {
```
nickeskov, line 13: Looks like a part of new message that can be sent by other nodes. You need to create new `EndorseBlockMessage` and support it in  `UnmarshalMessage` from `pkg/proto/proto.go`.

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+
+type FinalizationVoting struct {
+	EndorserIndexes                []int32        `json:"endorserIndexes"`
+	AggregatedEndorsementSignature []byte         `json:"aggregatedEndorsementSignature"`
+	ConflictEndorsements           []EndorseBlock `json:"conflictEndorsements"`
+}
+
+func (f *FinalizationVoting) Marshal() ([]byte, error) {
+	endBlockProto := f.ToProtobuf()
+	return endBlockProto.MarshalVTStrict()
```
nickeskov, line 111: Better to use `MarshalToProtobufDeterministic`.

pkg/proto/finalization.go
```diff
@@ -0,0 +1,108 @@
...
+	EndorserIndex        int32  `json:"endorserIndex"`
+	FinalizedBlockID     []byte `json:"finalizedBlockID"`
+	FinalizedBlockHeight uint32 `json:"finalizedBlockHeight"`
+	EndorsedBlockID      []byte `json:"endorsedBlockId"`
+	Signature            []byte `json:"signature"`
+}
+
+func (e *EndorseBlock) Marshal() ([]byte, error) {
+	endBlockProto := e.ToProtobuf()
+	return endBlockProto.MarshalVTStrict()
```
nickeskov, line 23: Better to use `MarshalToProtobufDeterministic`.

pkg/node/fsm/ng_state.go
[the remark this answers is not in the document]
esuwu: 1. This has been fixed
2. Yes, it's %v but it's for errors.Errorf, not fmt.Errorf. We have been using errors.Errorf pretty much everywhere in the project

pkg/miner/endorsementpool/endorsement_pool.go
esuwu: seq decides priority order,
index for consistent heap operations.

pkg/miner/endorsementpool/endorsement_pool.go
esuwu: Nope, just a defensive check since it's a pointer

pkg/node/fsm/ng_state.go
esuwu: always %v with errors.Errorf

pkg/node/fsm/ng_state.go
esuwu: No need to cache previous endorsement, they will be inside blocks

pkg/node/fsm/ng_state.go
esuwu: ?

pkg/state/state.go
esuwu: Maybe

pkg/proto/finalization.go
esuwu: There is no such method. VTStrict is determenistic

pkg/proto/finalization.go
esuwu: No such method

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,243 @@
...
+	"container/heap"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/wavesplatform/gowaves/pkg/crypto/bls"
+	"github.com/wavesplatform/gowaves/pkg/proto"
+)
+
+const endorsementPoolLimit = 128
```
nickeskov: Guess this should be taken from blockchain settings.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,243 @@
...
+			e.EndorserIndex, e.EndorsedBlockID.String())
+	}
+	if p.countLimit > 0 && len(p.byKey) >= p.countLimit {
+		return errors.New("the endorsement pool is full")
+	}
+
+	p.seq++
+	endorserPublicKey := p.endorsersPublicKeyCache.PublicKeyByEndorserIndex(e.EndorserIndex)
+	it := heapItemEndorsement{eb: e, seq: p.seq, endorserPK: endorserPublicKey}
+	heap.Push(&p.h, it)
```
nickeskov: `it` must be a pointer according to the implementation.

pkg/node/fsm/ng_state.go
```diff
@@ -219,8 +250,32 @@ func (a *NGState) MinedBlock(
...
+	}
+	if finalityActivated {
+		ok, vrfErr := a.baseInfo.endorsements.Verify()
+		if vrfErr != nil {
+			return a, nil, a.Errorf(vrfErr)
+		}
+		if !ok {
+			return a, nil, a.Errorf(errors.Errorf("endorsement verification failed at height %d", height))
+		}
+		canFinalize, finErr := a.baseInfo.storage.CalculateVotingFinalization(a.baseInfo.endorsements.GetEndorsers(), height,
```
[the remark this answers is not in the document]
esuwu: The current height is `height + 1`, check the caller function. So this height is correct and refers to the previous block, not current

pkg/node/fsm/ng_state.go
nickeskov: I mean you have to enrich mircoblock inside `MicroMiner` methods. FSM should receive complete microblock.

pkg/state/state.go
```diff
@@ -3191,3 +3192,48 @@ func (s *stateManager) Close() error {
...
+	// If endorsersBalance >= 2/3 totalGeneratingBalance
+	totalGenBalance := float64(totalGeneratingBalance)
+	endorsersGenBal := float64(endorsersGeneratingBalance)
+	ratio := endorsersGenBal / totalGenBalance
+
+	const twoThirds = 2.0 / 3.0
+	if ratio >= twoThirds {
+		return true, nil
+	}
+	return false, nil
```
nickeskov: Floats are not good for such comparisons. ([ref](https://0.30000000000000004.com/))
<details><summary>Suggestion</summary>
<p>

```diff
diff --git a/pkg/state/state.go b/pkg/state/state.go
index f462bd40..a0e1acb6 100644
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -3219,21 +3219,17 @@ func (s *stateManager) CalculateVotingFinalization(endorsers []proto.WavesAddres
 			return false, errors.Wrap(err, "endorsersGeneratingBalance overflow")
 		}
 	}
-	if totalGeneratingBalance == 0 {
+	if totalGeneratingBalance == 0 || endorsersGeneratingBalance == 0 {
 		return false, nil
 	}
-	if endorsersGeneratingBalance == 0 {
-		return false, nil
-	}
-
 	// If endorsersBalance >= 2/3 totalGeneratingBalance
-	totalGenBalance := float64(totalGeneratingBalance)
-	endorsersGenBal := float64(endorsersGeneratingBalance)
-	ratio := endorsersGenBal / totalGenBalance
-
-	const twoThirds = 2.0 / 3.0
-	if ratio >= twoThirds {
-		return true, nil
+	threeEndorserBalances, err := common.MulInt(3, endorsersGeneratingBalance)
+	if err != nil {
+		return false, errors.Wrap(err, "endorsersGeneratingBalance*3 overflow")
 	}
-	return false, nil
+	twoTotalBalances, err := common.MulInt(2, totalGeneratingBalance)
+	if err != nil {
+		return false, errors.Wrap(err, "totalGeneratingBalance*2 overflow")
+	}
+	return threeEndorserBalances > twoTotalBalances, nil
 }

```

</p>
</details> 

And, btw, you should use `>` instead of `>=`. From docs:
> With knowledge of the active generators’ balances, it becomes possible to compute, for each block, the ratio between the cumulative balance of generators that have voted for the block, i.e., those who have produced subsequent blocks on top of it, and the total active generating balance. Once this ratio **exceeds the threshold of 2/3**, the block is considered finalized.

**exceeds**, not reaches. Have you checked this in scala node source code?

pkg/state/state.go
nickeskov: Any actions?

pkg/settings/blockchain_settings.go
```diff
@@ -33,6 +33,8 @@ const (
 	lenWithDAOAndXTNBuybackAddresses = 2
 )
 
+const EndorsementPoolLimit = 128
```
nickeskov: Still should not be constant. See `FunctionalitySettings.MaxGenerators`.

pkg/proto/finalization.go
nickeskov: There is no such method, absolutely sure, it's function. Though, It's fine as is.

pkg/proto/finalization.go
nickeskov: Answered in similar issue.

pkg/proto/proto.go
```diff
@@ -81,6 +81,7 @@ const (
 	ContentIDMicroBlockSnapshotRequest PeerMessageID = 35
 	ContentIDBlockSnapshot             PeerMessageID = 36
 	ContentIDMicroBlockSnapshot        PeerMessageID = 37
+	ContentIDEndorseBlock              PeerMessageID = 38
```
nickeskov, line 84: Have you checked this value in scala node implementation?

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,244 @@
...
+	if n == 0 {
+		return false, errors.New("failed to verify endorsements, the endorsement pool is empty")
+	}
+
+	sigs := make([]bls.Signature, 0, n)
+	pks := make([]bls.PublicKey, 0, n)
+
+	for _, heapItem := range p.h {
+		var sig bls.Signature
+		if err := sig.UnmarshalJSON(heapItem.eb.Signature); err != nil {
```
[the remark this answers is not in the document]
nickeskov: I think it's valid issue.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,244 @@
...
+				heapItem.eb.EndorserIndex, err)
+		}
+		sigs = append(sigs, sig)
+		pks = append(pks, heapItem.endorserPK)
+	}
+
+	msg, err := p.h[0].eb.EndorsementMessage() // all endorsements use the same message
+	if err != nil {
+		return false, err
+	}
```
nickeskov: This is good assumption. But I thin message should be provided as an argument and constructed by the caller side. Trusting user data here can be an issue.

pkg/node/fsm/ng_state.go
```diff
@@ -205,6 +207,35 @@ func (a *NGState) Block(peer peer.Peer, block *proto.Block) (State, Async, error
...
+	endorsedBlockID := blockEndorsement.EndorsedBlockID
+	endorsedMicroBlock, found := a.baseInfo.MicroBlockCache.GetBlock(endorsedBlockID)
+	if !found || endorsedMicroBlock == nil {
+		return a, nil, a.Errorf(fmt.Errorf(
+			"failed to find the microblock that was endorsed (block ID: %s)",
+			endorsedBlockID.String(),
+		))
+	}
+	top := a.baseInfo.storage.TopBlock()
+	if top.BlockID() != endorsedMicroBlock.Reference { // Microblock's endorsement doesn't refer to last block
```
nickeskov: As I know, endorsements are sent for parent blocks. @alexeykiselev check me please, because I'm not sure about this statement.
It means that `top.BlockID` never equal `endorsedMicroBlock.Reference`, so the condition is always true.

pkg/node/fsm/ng_state.go
```diff
@@ -219,8 +250,32 @@ func (a *NGState) MinedBlock(
 	if heightErr != nil {
 		return a, nil, a.Errorf(heightErr)
 	}
+
+	finalityActivated, err := a.baseInfo.storage.IsActivated(int16(settings.DeterministicFinality))
+	if err != nil {
+		return a, nil, a.Errorf(err)
+	}
```
nickeskov: I think you have to check feature activation at new block height, not blockchain height. @alexeykiselev correct me if I'm wrong.

pkg/proto/block.go
```diff
@@ -321,15 +321,16 @@ type BlockHeader struct {
...
-
-	ID BlockID `json:"id"` // this field must be generated and set after Block unmarshalling
+	TransactionBlockLength uint32              `json:"transactionBlockLength,omitempty"`
+	TransactionCount       int                 `json:"transactionCount"`
+	GeneratorPublicKey     crypto.PublicKey    `json:"generatorPublicKey"`
+	BlockSignature         crypto.Signature    `json:"signature"`
+	TransactionsRoot       B58Bytes            `json:"transactionsRoot,omitempty"`
+	StateHash              *crypto.Digest      `json:"stateHash,omitempty"`        // is nil before protocol version 1.5
+	ChallengedHeader       *ChallengedHeader   `json:"challengedHeader,omitempty"` // is nil before protocol version 1.5
+	FinalizationVoting     *FinalizationVoting `json:"finalizationVoting,omitempty"`
```
nickeskov, line 350: Guess need comment when this field can be `nil` and when cannot.
Also, please, add safe getter for this field (I mean like `GetChallengedHeader` and `GetStateHash`)

pkg/proto/proto.go
```diff
@@ -2222,3 +2225,61 @@ func unmarshalBlockIDs(data []byte) ([]BlockID, error) {
 	}
 	return ids, nil
 }
+
+type EndorseBlockMessage struct {
```
nickeskov, line 2229: What about adding it to `createActions` function?

pkg/proto/protobuf_converters.go
```diff
@@ -1714,6 +1716,59 @@ func (c *ProtobufConverter) Block(block *g.Block) (Block, error) {
...
+
+func (c *ProtobufConverter) FinalizationVoting(finalizationVoting *g.FinalizationVoting) (FinalizationVoting, error) {
+	if finalizationVoting == nil {
+		return FinalizationVoting{}, errors.New("empty finalization voting")
+	}
+	conflictEndorsements := make([]EndorseBlock, 0, len(finalizationVoting.ConflictEndorsements))
+	for i, ce := range finalizationVoting.ConflictEndorsements {
+		if ce == nil {
+			continue
+		}
```
nickeskov, line 1786: Maybe return an error in this case? Or it's expected and possible situation?

pkg/state/state.go
```diff
@@ -3191,3 +3193,120 @@ func (s *stateManager) Close() error {
...
+
+// CommittedGenerators returns the list of Waves addresses of committed generators.
+func (s *stateManager) CommittedGenerators(periodStart uint32) ([]proto.WavesAddress, error) {
+	pks, err := s.stor.commitments.generators(periodStart)
+	if err != nil {
+		return nil, err
+	}
+	addresses := make([]proto.WavesAddress, len(pks))
+	for i, pk := range pks {
+		addr := proto.MustAddressFromPublicKey(s.settings.AddressSchemeCharacter, pk)
```
nickeskov: Better to use no `must` version.

pkg/state/state.go
```diff
@@ -3191,3 +3193,120 @@ func (s *stateManager) Close() error {
...
+
+	endPKb := endorserPK[:]
+	for _, cm := range rec.Commitments {
+		if bytes.Equal(endPKb, cm.EndorserPK[:]) {
+			return cm.GeneratorPK, nil
+		}
+	}
+
+	var empty crypto.PublicKey
+	return empty, fmt.Errorf("endorser public key not found in commitments for period %d", periodStart)
```
nickeskov: can be simplified to `crypto.PublicKey{}, ...`

pkg/state/state.go
```diff
@@ -3191,3 +3193,120 @@ func (s *stateManager) Close() error {
...
+		if errors.Is(err, keyvalue.ErrNotFound) {
+			return empty, fmt.Errorf("no commitments found for period %d", periodStart)
+		}
+		return empty, fmt.Errorf("failed to retrieve commitments record: %w", err)
+	}
+
+	var rec commitmentsRecord
+	if umErr := rec.unmarshalBinary(data); umErr != nil {
+		var empty crypto.PublicKey
+		return empty, fmt.Errorf("failed to unmarshal commitments record: %w", umErr)
```
nickeskov: the same

pkg/state/state.go
```diff
@@ -3191,3 +3193,120 @@ func (s *stateManager) Close() error {
...
+// FindGeneratorPKByEndorserPK finds the generator's Waves public key corresponding
+// to the given BLS endorser public key in the commitments record for the given period.
+func (s *stateManager) FindGeneratorPKByEndorserPK(periodStart uint32,
+	endorserPK bls.PublicKey) (crypto.PublicKey, error) {
+	key := commitmentKey{periodStart: periodStart}
+	data, err := s.stor.commitments.hs.newestTopEntryData(key.bytes())
+	if err != nil {
+		var empty crypto.PublicKey
+		if errors.Is(err, keyvalue.ErrNotFound) {
+			return empty, fmt.Errorf("no commitments found for period %d", periodStart)
```
nickeskov: maybe wrap this error?

pkg/state/state.go
```diff
@@ -3191,3 +3193,120 @@ func (s *stateManager) Close() error {
...
+
+// FindEndorserPKByIndex retrieves the BLS endorser public key by its index
+// in the commitments list for the given period.
+func (s *stateManager) FindEndorserPKByIndex(periodStart uint32, index int) (bls.PublicKey, error) {
+	key := commitmentKey{periodStart: periodStart}
+	data, err := s.stor.commitments.hs.newestTopEntryData(key.bytes())
+	if err != nil {
+		var empty bls.PublicKey
+		if errors.Is(err, keyvalue.ErrNotFound) {
+			return empty, fmt.Errorf("no commitments found for period %d", periodStart)
```
nickeskov: wrap error

pkg/state/state.go
```diff
@@ -3191,3 +3193,120 @@ func (s *stateManager) Close() error {
...
+
+	// If endorsersBalance >= 2/3 totalGeneratingBalance
+	totalGenBalance := float64(totalGeneratingBalance)
+	endorsersGenBal := float64(endorsersGeneratingBalance)
+	ratio := endorsersGenBal / totalGenBalance
+
+	const twoThirds = 2.0 / 3.0
+	if ratio > twoThirds {
+		return true, nil
+	}
```
nickeskov: scala implementation does the same? i remember i suggested example where only integers are used
i think  float numbers must not be used here.

pkg/state/commitments.go
```diff
@@ -110,3 +110,24 @@ func (c *commitments) size(periodStart uint32) (int, error) {
 	}
 	return len(rec.Commitments), nil
 }
+
+// size returns the number of commitments for the given period start.
+func (c *commitments) generators(periodStart uint32) ([]crypto.PublicKey, error) {
+	key := commitmentKey{periodStart: periodStart}
+	data, err := c.hs.newestTopEntryData(key.bytes())
+	if err != nil {
+		if errors.Is(err, keyvalue.ErrNotFound) {
```
nickeskov, line 363: If you're looking for data in history use `isNotFoundInHistoryOrDBErr` helper function.

pkg/proto/finalization.go
```diff
@@ -0,0 +1,106 @@
+package proto
+
+import (
+	"encoding/binary"
+
+	g "github.com/wavesplatform/gowaves/pkg/grpc/generated/waves"
+)
+
+const heightSize = 4
```
nickeskov: I think this constant should be local inside `EndorsementMessage` method. No need to do it package private. Also you can use `uint32Size` constant from `proto` package.

pkg/proto/block.go
```diff
@@ -450,6 +462,10 @@ func (b *BlockHeader) HeaderToProtobufHeader(scheme Scheme) (*g.Block_Header, er
 	if sh, present := b.GetStateHash(); present {
 		stateHash = sh.Bytes()
 	}
+	var finalizationVoting *g.FinalizationVoting
+	if b.FinalizationVoting != nil {
```
nickeskov: maybe better to use `GeFinalizationVoting` for similarity with other code?

pkg/proto/block.go
nickeskov: Please, add comment that this field is `nil` before protocol version 1.6.

pkg/node/fsm/ng_state.go
nickeskov: This is still valid issue. The block signature will be invalid with current code. 
<details><summary>Diff for this issue (should be checked)</summary>
<p>

```diff
diff --git a/pkg/grpc/protobuf-schemas b/pkg/grpc/protobuf-schemas
index 0a996a07..ac207a97 160000
--- a/pkg/grpc/protobuf-schemas
+++ b/pkg/grpc/protobuf-schemas
@@ -1 +1 @@
-Subproject commit 0a996a0762bbe28affb1c90fb8f2dcdb2e1cdad4
+Subproject commit ac207a97a883a4212c602a9898026aacf0d5820d
diff --git a/pkg/miner/micro_miner.go b/pkg/miner/micro_miner.go
index a6d4c5b5..21fcda3a 100644
--- a/pkg/miner/micro_miner.go
+++ b/pkg/miner/micro_miner.go
@@ -23,18 +23,20 @@ var ErrNoTransactions = errors.New("no transactions")
 var ErrStateChanged = errors.New("state changed")
 
 type MicroMiner struct {
-	state  state.State
-	utx    types.UtxPool
-	scheme proto.Scheme
-	logger *slog.Logger
+	state           state.State
+	utx             types.UtxPool
+	scheme          proto.Scheme
+	logger          *slog.Logger
+	endorsementPool types.EndorsementPool
 }
 
 func NewMicroMiner(services services.Services) *MicroMiner {
 	return &MicroMiner{
-		state:  services.State,
-		utx:    services.UtxPool,
-		scheme: services.Scheme,
-		logger: slog.Default().With(logging.NamespaceKey, "MICRO MINER"),
+		state:           services.State,
+		utx:             services.UtxPool,
+		scheme:          services.Scheme,
+		logger:          slog.Default().With(logging.NamespaceKey, "MICRO MINER"),
+		endorsementPool: services.EndorsementPool,
 	}
 }
 
@@ -240,6 +242,7 @@ func (a *MicroMiner) Micro(minedBlock *proto.Block, rest proto.MiningLimits, key
 		TotalResBlockSigField: newBlock.BlockSignature,
 		TotalBlockID:          newBlock.BlockID(),
 		StateHash:             sh,
+		Endorsements:          a.endorsementPool.GetAll(), // TODO: would it work correctly when finality is not active?
 	}
 
 	err = micro.Sign(a.scheme, sk)
diff --git a/pkg/node/fsm/ng_state.go b/pkg/node/fsm/ng_state.go
index d8ae0565..090ecf1a 100644
--- a/pkg/node/fsm/ng_state.go
+++ b/pkg/node/fsm/ng_state.go
@@ -423,9 +423,6 @@ func (a *NGState) mineMicro(
 	}
 	metrics.MicroBlockMined(micro, block.TransactionCount)
 
-	endorsements := a.baseInfo.endorsements.GetAll()
-	micro.Endorsements = endorsements
-
 	err = a.baseInfo.storage.Map(func(s state.NonThreadSafeState) error {
 		_, er := a.baseInfo.blocksApplier.ApplyMicro(s, block)
 		return er


```

</p>
</details> 

Also new block (after appending microblock) MUST contain new `FinalizationVoting` field which may be changed.

pkg/node/fsm/ng_state.go
```diff
@@ -205,6 +207,110 @@ func (a *NGState) Block(peer peer.Peer, block *proto.Block) (State, Async, error
...
+		return a, nil, proto.NewInfoMsg(err)
+	}
+
+	// TODO verify individual endorsements (signatures) using the endorser's public key.
+	activationHeight, actErr := a.baseInfo.storage.ActivationHeight(int16(settings.DeterministicFinality))
+	if actErr != nil {
+		return a, nil,
+			proto.NewInfoMsg(errors.Errorf("failed to get DeterministicFinality activation height, %v", actErr))
+	}
+	height, heightErr := a.baseInfo.storage.Height()
```
nickeskov, line 374: I think if current height is lower then activation height we should ignore such network messages or even close a connection with the sender.

pkg/node/fsm/ng_state.go
```diff
@@ -205,6 +207,110 @@ func (a *NGState) Block(peer peer.Peer, block *proto.Block) (State, Async, error
...
+	}
+
+	allEndorsers := a.baseInfo.endorsements.GetEndorsers()
+	endorsersAddresses := make([]proto.WavesAddress, 0, len(allEndorsers))
+	for _, endorser := range allEndorsers {
+		pk, findErr := a.baseInfo.storage.FindGeneratorPKByEndorserPK(periodStart, endorser)
+		if findErr != nil {
+			return findErr
+		}
+		addr := proto.MustAddressFromPublicKey(a.baseInfo.scheme, pk)
```
nickeskov: better to use no must version

pkg/node/fsm/ng_state.go
```diff
@@ -219,8 +325,19 @@ func (a *NGState) MinedBlock(
...
+
+	finalityActivated, err := a.baseInfo.storage.IsActiveAtHeight(int16(settings.DeterministicFinality), height+1)
+	if err != nil {
+		return a, nil, a.Errorf(err)
+	}
+	if finalityActivated {
+		if finErr := a.handleFinality(block, height); finErr != nil {
+			return a, nil, a.Errorf(finErr)
+		}
+	}
```
nickeskov: This should be placed in `MicroblockMiner.MineKeyBlock`. Also `proto.CreateBlock` function should be changed according with the new block parameter.

pkg/node/fsm/action.go
```diff
@@ -74,3 +75,25 @@ func (a *ActionsImpl) SendBlock(block *proto.Block) {
...
 }
+
+func (a *ActionsImpl) SendEndorseBlock(endorse *proto.EndorseBlock) {
+	bts, err := endorse.Marshal()
+	if err != nil {
+		a.logger.Error("Failed to marshal endorse block", logging.Error(err))
+		return
+	}
+
+	msg := &proto.EndorseBlockMessage{
```
nickeskov, line 86: Does this message have `Stringer`interface for user-readable representation in debug logging mode, for example?

pkg/node/actions_by_type.go
```diff
@@ -362,6 +362,27 @@ func MicroBlockSnapshotAction(
 	return fsm.MicroBlockSnapshot(mess.ID, blockID, blockSnapshot)
 }
 
+func EndorseBlockAction(
+	_ services.Services, mess peer.ProtoMessage, fsm *fsm.FSM, nl *slog.Logger,
+) (fsm.Async, error) {
+	protoMess := g.EndorseBlock{}
+	endorseMsg, ok := mess.Message.(*proto.EndorseBlockMessage)
+	if !ok {
+		nl.Debug("unexpected message type", slog.String("type", fmt.Sprintf("%T", mess.Message)))
```
nickeskov, line 380: Use `logging.Type` helper.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+// CleanAll safely resets the pool.
+func (p *EndorsementPool) CleanAll() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.byKey = make(map[key]*heapItemEndorsement)
+	p.items = nil
+	p.conflicts = nil
+}
+
+// Verify validates all endorsements in the pool by aggregating BLS signatures.
```
nickeskov: Duplicate

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+		var sig bls.Signature
+		if err := sig.UnmarshalJSON(it.eb.Signature); err != nil {
+			return false, fmt.Errorf("invalid signature at endorser index %d: %w",
+				it.eb.EndorserIndex, err)
+		}
+		sigs = append(sigs, sig)
+		pks = append(pks, it.endorserPK)
+	}
+
+	msg, err := p.items[0].eb.EndorsementMessage() // all endorsements use the same message
```
nickeskov: Have issue about this.

nickeskov: Have issue about this.

pkg/node/fsm/ng_state.go
nickeskov: Please, also add comment about why `height+1` is used.

pkg/state/state.go
nickeskov: Have you checked this?

pkg/node/fsm/ng_state.go
nickeskov: Then use `Wrap`/`Wrapf` 😃 
This holds info about internal error and its context.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+	byKey     map[key]*heapItemEndorsement
+	items     []*heapItemEndorsement // always sorted
+	conflicts []proto.EndorseBlock
+}
+
+func NewEndorsementPool() *EndorsementPool {
+	return &EndorsementPool{
+		byKey: make(map[key]*heapItemEndorsement),
+	}
+}
```
nickeskov: Does not see `limit` in endorsements. As I said it should be taken from `FunctionalitySettings`.
You can't return more than `MaxGenerators` endorsers by `GetAll` method.
Better to use `min-heap` for excluding the low-priority endorsers.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (p *EndorsementPool) insertSorted(item *heapItemEndorsement) {
+	i := sort.Search(len(p.items), func(i int) bool {
+		if p.items[i].balance != item.balance {
+			// descending balance: insert before smaller balances
+			return p.items[i].balance < item.balance
+		}
+		// ascending seq: insert before newer (larger seq)
+		return p.items[i].seq > item.seq
```
nickeskov: Can be simplified to one condition with `&&`.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	endorseBlocks := p.GetAll()
+	var finalizationVoting proto.FinalizationVoting
+	signatures := make([]bls.Signature, 0, len(endorseBlocks))
+	endorsersIndexes := make([]int32, 0, len(endorseBlocks))
+	for _, eb := range endorseBlocks {
+		var sig bls.Signature
+		if err := sig.UnmarshalJSON(eb.Signature); err != nil {
```
nickeskov: Why `UnmarshalJSON`???

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+		out = append(out, *it.eb)
+	}
+	return out
+}
+
+func (p *EndorsementPool) Finalize() (proto.FinalizationVoting, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	endorseBlocks := p.GetAll()
```
nickeskov: **This causes deadlock.**
<details><summary>Diff fix</summary>
<p>

```diff
diff --git a/pkg/miner/endorsementpool/endorsement_pool.go b/pkg/miner/endorsementpool/endorsement_pool.go
index 4e17a756..7c9e6a79 100644
--- a/pkg/miner/endorsementpool/endorsement_pool.go
+++ b/pkg/miner/endorsementpool/endorsement_pool.go
@@ -97,11 +97,12 @@ func (p *EndorsementPool) Finalize() (proto.FinalizationVoting, error) {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
-	endorseBlocks := p.GetAll()
-	var finalizationVoting proto.FinalizationVoting
-	signatures := make([]bls.Signature, 0, len(endorseBlocks))
-	endorsersIndexes := make([]int32, 0, len(endorseBlocks))
-	for _, eb := range endorseBlocks {
+	var (
+		signatures       = make([]bls.Signature, 0, len(p.items))
+		endorsersIndexes = make([]int32, 0, len(p.items))
+	)
+	for _, it := range p.items {
+		eb := it.eb
 		var sig bls.Signature
 		if err := sig.UnmarshalJSON(eb.Signature); err != nil {
 			continue
@@ -113,10 +114,11 @@ func (p *EndorsementPool) Finalize() (proto.FinalizationVoting, error) {
 	if err != nil {
 		return proto.FinalizationVoting{}, err
 	}
-	finalizationVoting.AggregatedEndorsementSignature = aggregatedSignature
-	finalizationVoting.EndorserIndexes = endorsersIndexes
-	finalizationVoting.ConflictEndorsements = p.conflicts
-	return finalizationVoting, nil
+	return proto.FinalizationVoting{
+		AggregatedEndorsementSignature: aggregatedSignature,
+		EndorserIndexes:                endorsersIndexes,
+		ConflictEndorsements:           p.conflicts,
+	}, nil
 }
 
 func (p *EndorsementPool) GetEndorsers() []bls.PublicKey {


```

</p>
</details>

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,194 @@
...
+// Verify validates all endorsements in the pool by aggregating BLS signatures.
+// Verify validates all endorsements in the pool by aggregating BLS signatures.
+func (p *EndorsementPool) Verify() (bool, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	n := len(p.items)
+	if n == 0 {
+		return false, errors.New("failed to verify endorsements: the endorsement pool is empty")
+	}
```
nickeskov, line 217: Maye this error should be package-public? or return no error if pool is empty?

pkg/state/state.go
esuwu: Integer division 2/3 will be 0, you can't calculate it correctly using integers

pkg/state/state.go
nickeskov: UPD: I'm wrong, `>=` is correct. Please, change back `>` to `>=`.

pkg/state/state.go
esuwu: Fixed

pkg/client/transactions.go
```diff
@@ -176,6 +177,7 @@ func (a *Transactions) Broadcast(ctx context.Context, transaction proto.Transact
 		}
 	}
 
+	slog.Debug("Broadcasting transaction", "transaction", string(bts))
```
nickeskov: Please, remove the log record

pkg/node/fsm/ng_state.go
nickeskov: Block endorsement must be ignored (according to the spec) before`DeterministicFinality` feature activation.

pkg/wallet/embedded_wallet.go
```diff
@@ -32,6 +34,25 @@ func (a *EmbeddedWalletImpl) SignTransactionWith(pk crypto.PublicKey, tx proto.T
 	return ErrPublicKeyNotFound
 }
 
+func (a *EmbeddedWalletImpl) KeyPairsBLS() ([]bls.PublicKey, []bls.SecretKey, error) {
+	seeds := a.seeder.AccountSeeds()
+	var publicKeys []bls.PublicKey
+	var secretKeys []bls.SecretKey
```
nickeskov: Better to allocate with `len(seeds)` capacity.

pkg/state/state.go
```diff
@@ -3354,6 +3356,112 @@ func (s *stateManager) Close() error {
...
+	return s.stor.commitments.newestExistsByEndorserPK(periodStart, endorserPK)
+}
+
+func (s *stateManager) NewestCommitedEndorsers(periodStart uint32) ([]bls.PublicKey, error) {
+	return s.stor.commitments.newestEndorsers(periodStart)
+}
+
+// CommittedGenerators returns the list of Waves addresses of committed generators.
+func (s *stateManager) CommittedGenerators(periodStart uint32) ([]proto.WavesAddress, error) {
+	return s.stor.commitments.CommittedGeneratorsAddresses(periodStart, s.settings.AddressSchemeCharacter)
```
nickeskov: I think this method should not use `newest` internally.

pkg/state/state.go
```diff
@@ -3354,6 +3356,112 @@ func (s *stateManager) Close() error {
...
+func (s *stateManager) LastFinalizedHeight() (proto.Height, error) {
+	height, err := s.stor.finalizations.newest()
+	if err == nil {
+		return height, nil
+	}
+	if !errors.Is(err, ErrNoFinalization) && !errors.Is(err, ErrNoFinalizationHistory) {
+		return 0, err
+	}
+	// No finalization found, calculate it.
+	return proto.CalculateLastFinalizedHeight(height), nil
```
nickeskov: Remind about auto-finalization for 100+ blocks behind.

pkg/miner/endorsementpool/endorsement_pool.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (h endorsementMinHeap) Len() int { return len(h) }
+
+func (h endorsementMinHeap) Less(i, j int) bool {
+	if h[i].balance == h[j].balance {
+		// Late (Higher seq), lower priority.
+		return h[i].seq > h[j].seq
+	}
+	// Lower balance, lower priority.
+	return h[i].balance < h[j].balance
```
nickeskov: can be simplified to `return h[i].balance < h[j].balance && h[i].seq > h[j].seq`

PR #2235, opened 2026-09-20
wmd: report symbol file scan errors

vividvisit:
Close symbol-list files after reading and propagate `bufio.Scanner` failures.

Previously an input error, including the scanner token-size limit, returned a partially populated symbol map without an error.


alexeykiselev:
Thank you for the contribution, but the `wmd` utility is outdated and will be removed soon.

